PR #31, opened 2021-03-25
Connect to controller regardless of authentication

walidkayhan:
Not sure if this is the solution to the issue as I was not able to reproduce the problem exactly. Though this does seem to make sense, there won't be a connection to the controller if the user is not authenticated so, maybe this is what is preventing the device list from populating.

### DIFF
--- a/src/app/index.jsx
+++ b/src/app/index.jsx
@@ -97,20 +97,17 @@ series([
         const token = store.get('session.token');
         user.signin({ token: token })
             .then(({ authenticated, token }) => {
-                if (authenticated) {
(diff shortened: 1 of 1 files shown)


kglovern:
I'm going to leave this unmerged for the time being - there's some behaviour issues that happen on build with it that will need to be looked at.

Specifically, where electron builds previously failed to load available devices, now it fails to load the page entirely leaving the application window empty with these changes which is probably a worse case scenario.

I'm not 100% what's causing this behaviour but its probably best we spend more time on it in the future

PR #95, opened 2022-06-30
MI-494 Custom Theme

sophiabeluli:
- added 2 new themes: custom light and custom dark
- colours of all theme parts changeable with a colour picker
- colour picker features a colour wheel, slider, preset colours, and manual hex and RGB input
- reset button allows return to default theme colours (light or dark)


walidkayhan:
Great work :) Everything seems to be fine with it! I added a small tweak to it to prevent theme changes when a file is loaded, figured this is the easiest solution to the bigger issue of the program running out of memory when switching themes when a large file is loaded

PR #96, opened 2022-07-06
Mi 667 custom decimal setting

kglovern:


kglovern:
@ssamal94 
Couple points of feedback:

- Labelling could be clearer - I'd probably change it to "precision" or "accuracy" as the control label over 10x since those are more colloquial terms
- Tooltip should be clearer - Make it clear that 0 disabled and uses 2 place for MM/3 for metric and any other value is used instead
- Validation - if you enter a value over the max, it should be set to the max (5) instead of 0.
- Validation - if you enter a value under 0 (-2) it should be set to 0

PR #97, opened 2022-07-07
Test Run Improvements

sophiabeluli:
- test run now uses the sender instead of the feeder
- added tool animation

### DIFF
--- a/src/app/widgets/Visualizer/GCodeVisualizer.js
+++ b/src/app/widgets/Visualizer/GCodeVisualizer.js
@@ -139,6 +139,10 @@ class GCodeVisualizer {
         this.frameIndex = frameIndex;
     }
 
+    getCurrentLocation() {
+        return this.vertices[this.frames[this.frameIndex].vertexIndex] ? this.vertices[this.frames[this.frameIndex].vertexIndex] : this.vertices[this.frames[this.frameIndex].vertexIndex - 1];
+    }
+
     unload() {
         this.geometry.dispose();
         this.group.clear();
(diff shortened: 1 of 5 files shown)


src/app/widgets/Visualizer/GCodeVisualizer.js
```diff
@@ -139,6 +139,10 @@ class GCodeVisualizer {
         this.frameIndex = frameIndex;
     }
 
+    getCurrentLocation() {
+        return this.vertices[this.frames[this.frameIndex].vertexIndex] ? this.vertices[this.frames[this.frameIndex].vertexIndex] : this.vertices[this.frames[this.frameIndex].vertexIndex - 1];
```
sophiabeluli, line 143: I have this null check here because for some reason, the last frame would point to a `vertexIndex` just outside the range of the vertices array. Not sure why this happens. I can look into it further if you want.

kglovern:
Validation tests:

Running job to completion normally should not put the firmware into check mode
Running job and stopping early should not put the firmware into check mode
Running "Test file" to completion should put the firmware into check mode, run the file, and exit check mode.
Running "Test file" and stopping early should put the firmware into check mode, and return it to idle (exit check mode) after stop is pressed
Running "Test File" and stopping early should put the firmware into idle (exit check mode), and should return the visualizer bit to the original wpos location

src/app/widgets/Visualizer/GCodeVisualizer.js
walidkayhan: That is weird, you could look into it if you'd like but if this check doesn't cause any issues or side effects then it's fine to leave it alone for now.

walidkayhan:
Looks great! The bit moves in super speed during the check which is awesome!

- I did notice that you are not able to resume the test run if you click pause in the middle of it. You could do one of two things to fix that:
    - Remove the pause button during a test run completely, 
    - Create a way to resume test runs. 
    - Want to say removing the pause button is a better and easier solution since I can't really think of a good reason to pause a test run
- The bit doesn't seem to be spinning during the test run

PR #101, opened 2022-07-11
Stored Job Counter

sophiabeluli:
in workplace settings:
- added job counter
- added tracker for time spent running


walidkayhan:
Looks good

- When you are running a job and click on the cancel job button immediately, it increments the time spent running as expected. But when you are running a job, then pause, then click the cancel job button, it does not increment like it is supposed to

PR #104, opened 2022-07-15
Mi 689 outline tool move to worker

ssamal94:
Workpiece outline task converted to a worker and moved from server to the frontend.


kglovern:
I made some fixes to this.  Please look for errors in the console when changing functions since that's generally indicative of some other issue.  In this case, the variable context was getting replaced with 500 since you were sending that using controller.command.

ssamal94:
Got it, I will make sure to check for console errors from now on. Also, thanks for the fix!

PR #107, opened 2022-07-18
Mi 696 log all grbl errors/alarms into electron log file

ssamal94:
Awaiting review


src/app/lib/controller.js
```diff
@@ -95,7 +94,8 @@ class Controller {
         'outline:start': [],
         'file:load': [],
         'file:unload': [],
-        'homing:flag': []
+        'homing:flag': [],
+        'electron-error': [],
```
kglovern, line 98: There is already an error listener setup, I'm not sure the reason we'd need to add another error listener.

You can probably refactor this code to use the 'error' event listener and the 'type' key can differentiate between what sort of error it is (firmware related, app related, etc.) - and remove this unique handler.

src/app/lib/controller.js
```diff
@@ -203,9 +206,12 @@ class Controller {
                     this.type = args[0];
                     this.state = { ...args[1] };
                 }
+                if (eventName === 'electron-error') {
+                    window.ipcRenderer.send('log-error', args[0]);
```
kglovern, line 210: You need to wrap this in an isElectron check - this will throw an error if users are accessing it from the headless view.  You can send the ipc message if it's electron, or console log/client log if it's web view.

src/main.js
```diff
@@ -159,6 +159,10 @@ const main = () => {
             });
 
             ipcMain.on('log-error', (channel, err) => {
+                if ('type' in err) {
+                    log.transports.file.level = 'error';
+                    log.transports.file.resolvePath = () => path.join(__dirname, '/logs.txt');
```
kglovern, line 166: I don't think the resolvepath is necessary, we use the default value.

src/server/controllers/Grbl/GrblController.js
```diff
@@ -107,6 +107,10 @@ class GrblController {
             this.ready = false;
             if (err) {
                 log.error(`Unexpected error while reading/writing serial port "${this.options.port}":`, err);
+                this.emit('electron-error', {
```
kglovern, line 110: OK - this is for a few changes but the ticket SPECIFICALLY SAID to log only GRBL alarms and errors, not application alarms and errors.  

You only needed to listen to those firmware messages and handle them, but instead added code for logging ALL application errors which is redundant since we already do this.

So, to clarify, GRBL errors and alarm messages (emitted by runner or feeder/sender) need to be handled in this unique manner, we already have logging in place for application errors.

src/server/controllers/Grbl/GrblController.js
```diff
@@ -625,6 +645,10 @@ class GrblController {
             if (alarm) {
                 // Grbl v1.1
                 this.emit('serialport:read', `ALARM:${code} (${alarm.message})`);
+                this.emit('electron-error', {
```
kglovern, line 648: This is one of the ones to keep - but I'd maybe make the type different to differentiate between GRBL_ERROR and GRBL_ALARM

walidkayhan:
Looks good

PR #108, opened 2022-07-20
Mi 707 improvements to patch notes feature

ssamal94:
Note: For 'About', all DIVs are now in percentage and it doesn't break anymore.

However, as discussed in the meeting, the modal size is hard coded to 1000px/700px and has to remain unchanged. One thing to note is that the modal doesn't scale when the screen size changes (ex: 3072px-1090px), the setting modal looks small because it has a fixed size and takes up only 1000px/700px no matter the screen size. Similarly, it looks huge for smaller screen sizes.


kglovern:
Looks good - I'd move the link to other patch notes outside of the text area as a permanent link outside the markdown area vs hiding it at the bottom.  Make sure you add the external link icon when you do - look at the "Help" section as an example of what icon I mean.

PR #109, opened 2022-07-21
Custom Theme Improvements

sophiabeluli:
- every component now has their name, hex and colour shown in a list
- tooltips added to explain how the new layout works
- how to choose colours:
	- click on the colour circle beside the component you want to trigger the pop up of the colour picker, which is now a modal
                    - click the "choose colour" button when you want to keep your current selection
	            - simply exit the modal if you do not want to keep the current selection
        	    - if you want to reset the colour to the default theme colour, press the reset button
	- to keep all your changes, press the "save" button


src/app/containers/Preferences/components/ColorCircle.js
```diff
@@ -0,0 +1,45 @@
...
+    useEffect(() => {
+        document.getElementById('colorButton' + index).style.backgroundColor = colour;
+    }, []);
+
+    pubsub.subscribe('colour:change', (msg, data) => {
+        const { currentPart, newColour } = data;
+        if (currentPart === part) {
+            document.getElementById('colorButton' + index).style.backgroundColor = newColour.hex ? newColour.hex : newColour;
+        }
+    });
```
walidkayhan, line 38: This pubsub subscription will be re-run on every re-render of this component, which is not ideal and might lead to some issues. You should also unsubscribe to them once the component is unmounted from the UI by using the pubsub.unsubscribe() function. I would place this subscription in the useEffect you already have setup in this component and unsubscribe from it in the cleanup function of the useEffect.

src/app/containers/Preferences/Preferences.jsx
```diff
@@ -570,36 +577,9 @@ class PreferencesPage extends PureComponent {
...
-                this.visualizerConfig.set(theme + ' Jogging Coordinate Lines', themeColours.joggingCoordinateLines);
-                this.visualizerConfig.set(theme + ' G0', themeColours.G0Color);
-                this.visualizerConfig.set(theme + ' G1', themeColours.G1Color);
-                this.visualizerConfig.set(theme + ' G2', themeColours.G2Color);
-                this.visualizerConfig.set(theme + ' G3', themeColours.G3Color);
-
-                pubsub.publish('theme:change');
+                return this.visualizerConfig.get(CUST_THEME + ' ' + part)
+                    ? this.visualizerConfig.get(CUST_THEME + ' ' + part)
+                    : defaultColour;
```
walidkayhan, line 582: The .get function actually takes a second argument that is a default value if it's not able to find the config from the first argument string you provided. It's a little cleaner to use that instead of the ternary operator you have here.

src/app/containers/Preferences/index.styl
```diff
@@ -477,4 +488,60 @@
...
+
+.dots {
+  overflow: hidden;
+}
+
+.dots:after {
+  content: '. . . . . . . . . . . . . . . . . . . . . . . . . . . . .';
+  color: #000;
+  white-space: nowrap;
+}
```
walidkayhan, line 534: Think this is causing a UI bug where the container of the preferences is breaking out of the modal container. There are two easier ways to handle this, could not have these dots and just add more spacing between each color picker section, or you could replace this with a line div where you create a div and set its background and set its height to something small so it looks like a line.

src/app/widgets/Visualizer/index.jsx
```diff
@@ -964,36 +963,20 @@ class VisualizerWidget extends PureComponent {
...
+                zAxisColor: this.config.get('Custom Z Axis') ? this.config.get('Custom Z Axis') : DARK_THEME_VALUES.zAxisColor,
+                limitColor: this.config.get('Custom Limit') ? this.config.get('Custom Limit') : DARK_THEME_VALUES.limitColor,
+                cuttingCoordinateLines: this.config.get('Custom Cutting Coordinate Lines') ? this.config.get('Custom Cutting Coordinate Lines') : DARK_THEME_VALUES.cuttingCoordinateLines,
+                joggingCoordinateLines: this.config.get('Custom Jogging Coordinate Lines') ? this.config.get('Custom Jogging Coordinate Lines') : DARK_THEME_VALUES.joggingCoordinateLines,
+                G0Color: this.config.get('Custom G0') ? this.config.get('Custom G0') : DARK_THEME_VALUES.G0Color,
+                G1Color: this.config.get('Custom G1-3') ? this.config.get('Custom G1-3') : DARK_THEME_VALUES.G1Color,
+                G2Color: this.config.get('Custom G1-3') ? this.config.get('Custom G1-3') : DARK_THEME_VALUES.G2Color,
+                G3Color: this.config.get('Custom G1-3') ? this.config.get('Custom G1-3') : DARK_THEME_VALUES.G3Color,
```
walidkayhan, line 979: Like i mentioned in another comment, you can pass a default value as the second argument in .get instead of using the ternary operator. Just makes the code more maintainable and more clean

src/app/containers/Preferences/Visualizer/Theme.js
```diff
@@ -57,63 +68,90 @@ const Theme = ({ state, actions }) => {
...
+                                parts.map((value, i) => {
+                                    return (
+                                        <div key={i} className={styles.colorContainer}>
+                                            <span className={styles.first}>{value === G1_PART ? 'G1-G3' : value}</span>
+                                            <div className={styles.dots}></div>
+                                            <div className={styles.colorDisplay}>
+                                                <span>{themeColours.get(value)}</span>
+                                                <ColorCircle part={value} onClick={openModal} colour={themeColours.get(value)} index={i}/>
+                                            </div>
+                                        </div>
```
walidkayhan, line 141: Maybe making the whole section clickable would be easier for the user since I have a feeling some people will complain that they are having a hard time selecting the color circle

walidkayhan (commented):
Left a few comments on some areas in your code. I did notice the container breaks out of the modal at larger screen sizes due to the dots being too long

Screen Shot 2022-07-22 at 9 25 18 AM

walidkayhan:
Nice work, though i now notice the cutting and jogging coordinate lines options are breaking out of the fieldset on larger screen sizes. It's obviously due to the number of characters in the label, you could maybe try to decrease the font size on them, or if you have a better in mind by all means

Screen Shot 2022-07-22 at 1 49 39 PM

sophiabeluli:
> 

I changed those labels to use "coord" instead of the full word. That should make them a good length on all screen sizes/resolutions.

PR #110, opened 2022-07-21
Safety Preferences

sophiabeluli:
- added safety page to preferences: includes workspace warnings, soft limit warning, and safe height
- fixed bug in visualizer options where it tried to pull eeprom $20 before machine was connected


walidkayhan:
Looks great!

PR #111, opened 2022-07-21
Job Status/ Override toggle added

ssamal94:


walidkayhan:
Looks great, I noticed that if you are not connected to a machine and you load a file, you will get undefined for the feedrate percentage. You could do two things, don't show the toggle overrides when you are not connected to a machine, or set a default value for the feedrate percentage, like 0%.

Screen Shot 2022-07-22 at 10 55 08 AM

ssamal94:
Bug fixed: The default value is set to 0.

walidkayhan:
# Issue
The program actually crashes when you haven't connected before and toggle over to the overrides section

## Steps to reproduce
- Disconnect and make sure you don't have re-connect automatically toggled on
- Reload the page
- Load a file
- Toggle over to overrides

Screen Shot 2022-07-22 at 11 53 25 AM

PR #113, opened 2022-07-25
Spindle Speed Slider

sophiabeluli:
- when spindle is on, send updated speed command after slider move


walidkayhan:
Looks good, debounce seems to be working as expected.

When the laser is on and I toggle back to spindle mode, I get this error for some reason:
Screen Shot 2022-07-25 at 3 45 34 PM

sophiabeluli:
> Looks good, debounce seems to be working as expected.
> 
> When the laser is on and I toggle back to spindle mode, I get this error for some reason: Screen Shot 2022-07-25 at 3 45 34 PM

For some reason there was a function call for a function that doesn't exist, so I commented it out. I also made a couple fixes for some more bugs I found.

PR #115, opened 2022-07-29
scroll fixes

ssamal94:


walidkayhan:
Great work, fixed the UI bugs that were mentioned in the issue. I noticed a small thing, there seems to be a horizontal scrollbar showing up at the bottom of the table now that doesn't seem to do anything as well for some reason, and it doesn't seem like it should be there since there shouldn't be anything overflowing horizontally. Would be best to get rid of it somehow:

Screen Shot 2022-07-29 at 4 16 23 PM

kglovern:
The table no longer shows the actual key shortcut, but instead shows the label twice?
binds

ssamal94:
Column data is now fixed.

PR #116, opened 2022-08-01
Visualize SVG in Lite Mode

sophiabeluli:
**Summary of Changes:**
- lite mode's visualizer is now an svg
- this svg is:
	- a 2d image
	- rendered in top down view
	- centered in the visualizer area, with no grid
- this svg is also used in the surfacing tab when lite mode is enabled

**Current Issues:**
- the references are not fully working yet, so some actions like closing a file may not work as expected

### DIFF
--- a/src/app/widgets/Visualizer/PrimaryVisualizer.jsx
+++ b/src/app/widgets/Visualizer/PrimaryVisualizer.jsx
@@ -86,7 +86,6 @@ const PrimaryVisualizer = ({ actions, state, capable, showLoading, showRendering
                             state={state}
                             actions={actions}
                             containerID={containerID}
-                            liteMode={liteMode}
                             isSecondary={false}
                         />
                         <WorkflowControl
@@ -94,7 +93,6 @@ const PrimaryVisualizer = ({ actions, state, capable, showLoading, showRendering
                             state={state}
                             actions={actions}
                             invalidGcode={invalidLine.line}
-                            liteMode={liteMode}
                         />
 
 
--- a/src/app/widgets/Visualizer/PrimaryVisualizer.jsx
+++ b/src/app/widgets/Visualizer/PrimaryVisualizer.jsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
@@ -29,8 +29,6 @@ const PrimaryVisualizer = ({ actions, state, capable, showLoading, showRendering
     const { handleLiteModeToggle, handleRun, reset } = actions;
 
     const containerID = 'visualizer_container';
-
-    let visualizer = useRef();
 
     return (
         <Widget className={styles.vizWidgetOverride}>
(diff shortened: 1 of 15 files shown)


walidkayhan:
Awesome, have a few initial notes before i look more into the code:

- Allow user to opt-in to enable SVG visualization instead of it being enabled when in litemode (this will be another setting in the visualizer preferences area)
- Having the same background colour as the regular three.js visualizer would make the SVG mode look better and the transition between the two will look less daunting
- The SVG mode seems to cause the widget to break out of its container and take more horizontal space than it should:

https://user-images.githubusercontent.com/45078505/182912298-9f3712da-7e90-4b22-abb2-6e63580cd8fc.mov

src/app/widgets/Visualizer/PrimaryVisualizer.jsx
```diff
@@ -30,6 +30,8 @@ const PrimaryVisualizer = ({ actions, state, capable, showLoading, showRendering
 
     const containerID = 'visualizer_container';
 
+    let visualizer = useRef();
+
```
walidkayhan, line 34: No need to use a ref here, see comment below for more info

src/app/widgets/Visualizer/index.jsx
```diff
@@ -1128,6 +1128,7 @@ class VisualizerWidget extends PureComponent {
                     showVisualizer={showVisualizer}
                     visualizerRef={(ref) => {
                         this.visualizer = ref;
+                        console.log(ref);
```
walidkayhan, line 1131: Instead of setting **this.visualizer** to the exact ref of the visualizer, you can just set it to the visualizer property of the ref, since it is present as a ref in the visualizer wrapper component you set up and is being passed to the primary/secondary visualizer components. ( ex. **this.visualizer = ref?.visualizer**, we add optional chaining here so we don't run into problems when the ref has not been set yet and doesn't have a visualizer property )

src/app/widgets/Visualizer/PrimaryVisualizer.jsx
```diff
@@ -77,15 +79,15 @@ const PrimaryVisualizer = ({ actions, state, capable, showLoading, showRendering
                             state={state}
                             actions={actions}
                         />
-                        <Visualizer
+                        <VisualizerWrapper
                             show={showVisualizer}
                             cameraPosition={cameraPosition}
-                            ref={visualizerRef}
+                            ref={visualizer}
```
walidkayhan, line 85: Can just pass the visualizerRef prop here instead ( ex. **ref={visualizerRef}** )

src/app/widgets/Visualizer/SecondaryVisualizer.jsx
```diff
@@ -1,29 +1,30 @@
...
-                    ref={visualizerRef}
+                    ref={visualizer}
                     state={state}
                     actions={actions}
                     surfacingData={surfacingData}
                     containerID={SURFACING_VISUALIZER_CONTAINER_ID}
+                    isSecondary={true}
                 />
             )}
         </div>
```
walidkayhan, line 30: Just need to do the same as the comments I mentioned in PrimaryVisualizer.jsx

walidkayhan (commented):
Looks great, seems to work smoothly with just the issue with the references not working correctly and hence some function calls in the index.jsx file do not work as they are trying to call functions within the ref

PR #117, opened 2022-08-02
Mi 697 add component to safety section to view list of past grbl errors/alarms

ssamal94:
I can't start Electron. Please verify my file read logic in 'electronErrors:fetch' event inside GrblController.js to see if it is able to grab the correct file to read errors/alarms from. Thanks!


ssamal94:
NOTE: 
File - src/app/containers/Preferences/Safety/index.jsx
Line - 51

Please change 'isElectron()' to (not isElectron) !isElectron() to test this component on web app.

ssamal94:
**Changes and Things tested:**

- Error component should only show up in electron
- No console errors on web and electron
- No server errors
- Current styling does not affect any other component (all new styles are inside wrapper)
- Error list sorted based on the time stamp, latest error/alarm comes on top.
- Error list capped to 50 latest errors/alarms.
- User sees a message if no history of errors/alarms.
- Error list shows even if no machine is connected.
- Ux changes as per suggestion

kglovern:
This is getting merged in, but will have some tasks related to it pop up in the next sprint.

PR #137, opened 2022-08-15
Datafilter Issues

sophiabeluli:
- edited bracket comment regex
- removed u/FEFF character

### DIFF
--- a/src/server/controllers/Grbl/GrblController.js
+++ b/src/server/controllers/Grbl/GrblController.js
@@ -230,7 +230,7 @@ class GrblController {
                 let commentMatcher = /\s*;.*/g;
                 let comment = line.match(commentMatcher);
                 const commentString = (comment && comment[0].length > 0) ? comment[0].trim().replace(';', '') : '';
-                line = line.replace(commentMatcher, '').trim();
+                line = line.replace(commentMatcher, '').replace('/uFEFF', '').trim();
                 context = this.populateContext(context);
 
                 if (line[0] === '%') {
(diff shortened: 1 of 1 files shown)


src/server/controllers/Grbl/GrblController.js
```diff
@@ -342,12 +342,13 @@ class GrblController {
             dataFilter: (line, context) => {
                 // Remove comments that start with a semicolon `;`
                 let commentMatcher = /\s*;.*/g;
-                let bracketCommentLine = /^\s*\([^\)]*\)/gm;
+                let bracketCommentLine = /\([^\)]*\)/gm;
                 let toolCommand = /(T)(-?\d*\.?\d+\.?)/;
                 line = line.replace(bracketCommentLine, '').trim();
                 let comment = line.match(commentMatcher);
                 let commentString = (comment && comment[0].length > 0) ? comment[0].trim().replace(';', '') : '';
                 line = line.replace(commentMatcher, '').trim();
```
kglovern, line 350: Could you please chain the two replace calls and have a single trim rather than do trim twice?

src/server/controllers/Grbl/GrblController.js
```diff
@@ -342,12 +342,13 @@ class GrblController {
...
+                line = line.replace('/uFEFF', '').trim();
```
kglovern, line 351: Could you also apply this change to the feeder datafilter

sophiabeluli: done!

src/server/controllers/Grbl/GrblController.js
sophiabeluli: done!

PR #145, opened 2022-08-23
Mi 724 set wcs zero on reconnect if there are no limit switches

ssamal94:


kglovern:
1) Why are you storing the preference in local storage?  This won't work in the electron application.  Use the store library if it's a setting since that already handles differentiating local storage vs actual file for electron vs web view.
los_1

2)  It doesn't look like this localStorage value is referenced at all in the actual workflow

3)  Why does the setting update make an API call?  Why is this not just passed an as option on connection like we previously discussed?

4) Z zero isn't reset.

5)  Workspace changes are stored on workspace dropdown - what happens when the workspace is changed over a toolpath?

6)  Too much extraneous logging  and debug statements remain.

It looks like you had two conflicting ways to approach this task and didn't really complete either of them.

ssamal94:
I will try to fix most of the issues with all the pulls today.

ssamal94:
All required changes are pushed

PR #151, opened 2022-08-25
Fix jogging

cotepat:
Continuous jogging is now working when homing is BACK_RIGHT (homingFlagSet is false) and soft limits are enabled ($21=1). Also added an OFFSET in that case (just like the offset in the other homing cases)


cotepat:
The problem is that the parameter "direction" is not -1 or 1, it reflects the jogging step that is currently selected (for instead +5 or -5 mm). Thus the calculation is wrong, and the condition "direction === 1" is never met!

PR #157, opened 2022-08-26
Mi 759 delay machine connection status until grbl instance is created

ssamal94:
**Changes made:**

- User gets a feedback when GRBL is loading. Machine connection says ‘Loading firmware...'

- Machine disconnects with toast message if GRBL is not ready in 5 seconds

Toast message: 'Unable to fetch firmware details in time'


kglovern:
As mentioned in discussions "Loading" is the incorrect term since we're not loading anything, it's "Connecting" which should be the term used.

Forcing disconnect is also not functionality we want - we want to alert the user (error in machine status area, toast as well) but keep the connection active and just inform the user.

ssamal94:
Done!

src/app/store/defaultState.js
```diff
@@ -133,6 +133,7 @@ const defaultState = {
             }
         ],
         recentFiles: [],
+        grblExists: true,
```
kglovern, line 136: Why is this being stored in store and not redux?  It doesn't need to persist since it's a state check on a per connection basis, not a setting that needs to be saved on the hard drive.

src/app/widgets/NavbarConnection/NavbarConnection.jsx
```diff
@@ -47,7 +47,7 @@ class NavbarConnection extends PureComponent {
         if (connected && isControllerReady) {
             return 'Connected';
         } else if (connected && !isControllerReady) {
-            return 'Loading firmware...';
+            return 'Connecting...';
```
kglovern, line 50: Remove ellipsis

ssamal94:
All requested changes have been pushed.
Bug fixed: Machine status shows 'disconnected' while connecting instead of 'Invalid firmware response'.

src/app/widgets/NavbarConnection/index.jsx
```diff
@@ -390,6 +390,7 @@ class NavbarConnectionWidget extends PureComponent {
         const actions = {
             ...this.actions
         };
+        store.set('grblExists', true);
```
kglovern, line 393: This is still using the store and not Redux like originally asked?

kglovern:
I'm a little confused where you actually check whether the firmware exists.

As far as I can tell:

inside initController, we send "$$" to fetch settings - and you immediately send a "grbl:ready" event.  At no point do you check to see if the firmware sent a response.

"grbl:ready" is listened to inside controllerSagas and ALWAYS sends true along with the "grblExists:update".  Where is it possible to send "false" as a value?  So at this point it both a) never checks for a response and b) can never send anything but "true" as a value.

This means the check timeout will always evaluate to true and notification is never sent.

The goal was to verify whether the device we (successfully) connect to actually contains a firmware instance and update the machine status area accordingly instead of leaving it saying "disconnected" while the connection area says "connected".  Where is this check actually run - since as far as I can tell there is no logic to do so.

PR #165, opened 2022-09-07
Mi 762 main   improvements to flashing ux

ssamal94:


src/app/widgets/NavbarConnection/index.jsx
```diff
@@ -366,6 +367,7 @@ class NavbarConnectionWidget extends PureComponent {
         const actions = {
             ...this.actions
         };
+        store.set('ports', ports);
```
kglovern, line 370: Why is this being stored in the store?  The store is for application preferences, not transient data.  Is there a reason you're not getting the list of port from controller.ports?

kglovern:
- The port should be prepoulated with the connected port if connected before flashing
- The machine profile should be prepopulated with the selected machine profile in the main firmware window
- The incorrect information is getting sent to the backend since selecting a MKII machine profile is flashing with a MKI firmware as shown in the startup string.

kglovern:
Com port list is unpopulated if the device is disconnected when flash GRBL is clicked, which means the user cannot select a port at all - this is a regression of behaviour from the last time we looked at it.

flashing_1

Com Port list is also unpopulated in situations where the device is connected but there might be other options the user wishes to select.  Can still send to the port the user is connected to.

Selecting a different machine profile inside the window should save the machine profile to match that - as an example, if I had Generic CNC selected in the main firmware window, but inside Flash Grbl selected 30X30 MKII, the next time I open the firmware window Generic CNC is still selected.  The saved "machine profile" should match what they actively choose to flash.

kglovern:
- Duplicate styled-components dependency inside package.json - remove the out of date one.

- This **STILL** doesn't work and throws an (unreadable error) when connected to a port and occasionally when not connected.

Steps to recreate: 

1) Connect to microcontroller
2) Open firmware window
3) Click "Flash GRBL"
4) Verify the port is the same as you're connected to
5) Application throws error, flashing doesn't continue.  Error message consists of an object with no readable text.

error_1
error_2


Related: Attempting to flash after the above error will result in even more server errors since the controller no longer is connected and the close call is invalid.  You need some guards somewhere in your logic
error_3
.

ssamal94:
### Things tested:
- Flash with no machine connected on app start. (select machine from flash port dropdown). - working
- Connect machine after app launch and flash - working
- Open flash modal, unplug machine(USB), connect back; flash - working
- Change machine profile and flash, hit restore default, retains the new default machine profile
- Delete all kinds of storages from chrome and start app, do not connect machine from workspace widget, check firmware modal. Won’t let you click ‘Connect on last active port’. Has a message for user. 
- Delete all kinds of storages from chrome and start app; try to connect the machine from workspace widget. Works good
- Connect to machine, disconnect (from workspace widget), check firmware modal again. ‘Connect on last active port’ is now enabled.
- Start app -> connect machine -> flash -> do not connect machine, try to flash this time with selecting port form flash modal - works fine.  

### Any error or warnings:
- No errors (The machine was not closed properly error is gone with unsetting the store)

PR #167, opened 2022-09-13
Pop Out Window Functionality

sophiabeluli:
- added the ability to pop out the console window by clicking an icon in the corner of the console widget
	- this new window is on a new route: `<base_url>/#/widget/<widget_name>`
- new window can send console commands to the software, and those commands are also printed in the console in the main window
	- this was implemented by adding the new socket to the controller
- uses ipc events to get information from the main window to the pop out window (state of the widget, controller port)

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,6 +156,8 @@
     "@fortawesome/fontawesome-free": "^5.15.2",
     "@fortawesome/fontawesome-free-brands": "^5.0.13",
     "@fortawesome/free-regular-svg-icons": "^5.15.2",
+    "@mdi/js": "^7.0.96",
+    "@mdi/react": "^1.6.1",
     "@mui/icons-material": "^5.8.4",
     "@mui/material": "^5.9.2",
     "@mui/styled-engine-sc": "^5.8.0",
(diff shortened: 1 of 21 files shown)


walidkayhan:
- Popped out console cannot re-connect to a device if disconnected
- Styling seems a bit off, can expand/shrink the console window and the bottom action bar with the input, copy and run buttons do not move down to the end of the window as would expect it to do. You can probably set a min and max window size specifically for these console windows but will still need to adjust the styling for the action bar

Screen Shot 2022-09-16 at 10 09 01 AM

Screen Shot 2022-09-16 at 10 09 44 AM

sophiabeluli:
> 
done!

walidkayhan:
- Input area seems a bit large when expanding the screen size of the popped-out console, believe it has something to do with the grid-template-columns on the wrapper div that holds it and the console area, would probably need to adjust it to have a maximum height maybe
Screen Shot 2022-10-14 at 10 33 52 AM

- Can shrink the popped-out console screen to a very small size, should put a min screen size for it to now allow that, id maybe set it as the default size it starts off with it seems like a minimum size
Screen Shot 2022-10-14 at 10 38 20 AM

PR #176, opened 2022-09-30
widgets style changes to remove scroll on zoomed view

ssamal94:


kglovern:
These changes fundamentally changes how the app looks at various resolutions and window sizes and are largely outside the scope of the original task.   A very large number of the styling restrictions and sizing you removed were there for a reason.

The web view is your changes and the application is current behaviour on wide screen.

1 13_wide
changes_wide

Original vs your changes on 1080p. 

window_small
window_changes

The differences should be obvious to you - spacing and sizing on the widget column is much higher.  While it does address the scroll on scaled resolutions, it's has too many side effects on the look and feel of the application to be merged.

ssamal94:
The reason we were getting a scroll is because we used grids on the widget section. Grid contents do not shrink when the screen resolution changes, it results a scroll. I will try some other way out to match the existing scale of the app on different screen resolutions.

PR #181, opened 2022-10-07
Mi 779 add colour coded output for errors, alarms, and ok messages in the console

ssamal94:
### Changes:

1. error - **xterm(196)** (red)
2. response from the server/ OK Messages -  **default** (white)
3. User wrote something - **xterm(249)** (grey)
4. Alarms - **xterm( 167)** (variant of red)

To simulate alarm, add the following:

```
const testAlarm = 'ALARM: Probe fail';
            this.runner.parse(testAlarm);
```
at line 713 of GrblController.js (at the end of event ‘startup’).

Screen Shot 2022-10-06 at 7 56 21 PM

Screen Shot 2022-10-06 at 7 57 50 PM

### DIFF
--- a/src/app/widgets/Console/Terminal.jsx
+++ b/src/app/widgets/Console/Terminal.jsx
@@ -223,9 +223,9 @@ class TerminalWrapper extends PureComponent {
 
     writeln(data) {
         this.term.write('\r');
-        if (data.toLowerCase().includes('error')) {
+        if (data.includes('error:')) {
             this.term.write(color.xterm(RED)(data));
-        } else if (data.toLowerCase().includes('alarm')) {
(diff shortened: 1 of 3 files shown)


src/app/widgets/Console/Terminal.jsx
```diff
@@ -220,7 +223,13 @@ class TerminalWrapper extends PureComponent {
 
     writeln(data) {
         this.term.write('\r');
-        this.term.write(data);
+        if (data.toLowerCase().includes('error')) {
```
kglovern, line 226: Do you need the double toLowerCase calls here?  ALARM: and error: are constant responses from the firmware and will never have alternative capitalization.  

You can also include the colon after each.

PR #191, opened 2022-10-31
$C Crashes When File Loaded

sophiabeluli:
- fixed error where going into check mode with file loaded would crash program


kglovern:
I made some changes to this so it works as expected (animates bit again) - getCurrentLocation should have returned an object, and the offset calculation was a bit off.

@walidkayhan when you get a chance could you verify this is working as expected for both of us ($C in console no longer crashes app, Test Run properly tests file and animates bit) and merge when done.

Thanks!

PR #198, opened 2022-11-03
Mobile Style Improvements

sophiabeluli:
- changed the way mobile/tablet state is calculated: now checks for the ratio instead of the pixel count
- in mobile layout: 
    - connection bar is full width and expands to fit 
    - widgets are full width and expand to fit 
    - widget grid has static height but allows user to scroll


kglovern:
I get the same result as previous running this branch in remote mode accessing from an Iphone 13 - tablet warning always appears, visualizer appears in phone view

PR #199, opened 2022-11-07
Probe Offset Incorrect When $13 Enabled

sophiabeluli:
- fixed incorrect XYZ offsets when EEPROM 13 enabled and metric preferred units


kglovern:
This was also merged into the probe rework so I'm closing this request since it'll be merged with that

PR #211, opened 2022-11-15
Laser Custom Theme

sophiabeluli:
- added laser colour
- added support for more custom themes, making it easier to add new ones in the future 
     - in `constants.js`, create the constant for the new theme and add it to two constant arrays, `ALL_THEMES` and `CUSTOMIZABLE_THEMES`
     - that's all!
- changed the way the theme colours are stored, making it easier to add new ones in the future
     - in constants.js, create a new part constant and add it to `PARTS_LIST` and `DARK_THEME_VALUES`/`LIGHT_THEME_VALUES`
- now recreates cutting pointer on theme change so that its colour updates
- fixed "duplicate key" errors when opening settings

### DIFF
--- a/src/app/containers/Preferences/Preferences.jsx
+++ b/src/app/containers/Preferences/Preferences.jsx
@@ -44,7 +44,7 @@ import styles from './index.styl';
 import { METRIC_UNITS, WORKFLOW_STATE_RUNNING } from '../../constants';
 import { convertToImperial, convertToMetric } from './calculate';
 import {
-    CUST_THEME, DARK_THEME_VALUES,
+    DARK_THEME_VALUES,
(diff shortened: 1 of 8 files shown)


kglovern:
I think there was a bit of an misunderstanding what was asked for here - we're fine with the single custom theme, we wanted to be able to customize the colour used for laser visualization when laser mode was enabled.  The original title was "add default laser colour to theme and make it customizable" - it might not have been clear that it was adding an option TO existing themes, not adding a new laser theme.

So similar to how G0, G1-3 have their own colours, Laser should have a colour - by default it's currently hard coded to a red.

This is adding an additional theme completely which is a little beyond what we want.

kglovern:
laser

PR #217, opened 2022-11-18
Add Start Event to Start From Line

sophiabeluli:
- added start event to startFromLine if it exists


kglovern:
This was worded to prepend for a reason - can you grab the event code if enabled and just prepend it to modalGcode block rather than use the event in this case?  There is some strange behaviour when using feeder callback into feeder callback that we're trying to avoid.

PR #223, opened 2022-11-22
Change Z Probe Image

sophiabeluli:


kglovern:
We need the original image back.

The new image shows up if Z-only probe selected (Preferences -> Probe -> Touchplate type -> Z Probe) - Z only probe are generally just the circular block that you touch off the top of.

The other image is used when Z probing of a regular touchplate - there is a separate image for auto.

So Z Probe -> Z-Axis = use new image
Standard touchplate -> Z axis probe -> show previous/deleted image with it touching off the top or back of block of the square
Auto diameter touchplate - Z axis Probe -> Separate image of it touching off that profile

sophiabeluli:
> 

fixed

PR #224, opened 2022-11-23
Recovery on Machine Disconnect

sophiabeluli:
- added pop up when machine disconnects without user input during a job
    - allows the user to reconnect to the last active port and start from the last line

### DIFF
--- a/src/app/containers/Firmware/utils/index.js
+++ b/src/app/containers/Firmware/utils/index.js
@@ -21,7 +21,7 @@ export const controllerSettingsLoaded = () => {
 };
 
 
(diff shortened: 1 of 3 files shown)


kglovern:
Looks mostly good.

I'd probably reword the text to more clearly indicate that it looks like the connection was disrupted - right now the title is the only indication of what happened and it might be 100% clear.

The text also isn't explicit that  "Resume" will attempt to reconnect to the last active port and resume.  It reads like it's expecting the user to connect and then hit resume which is not possible due to the modal state.  I'd probably find a way to make it clear that that button will attempt to reconnect and resume sending.

image

PR #233, opened 2022-11-30
EventTrigger Refactor

sophiabeluli:
- configStore now stores events as a map
	- event name => event object (ex. `gcode:start => { enabled: true, commands: 'X0', .... }`)
	- there are constants for each of these names to increase maintainability
- migration has been written to convert old format .sender_rc (array of events) to new format (events object)
- since events are in a map and there can only be one event per key, therefore EventTrigger has been rewritten to reflect that
	- this includes more general functions where the user passes a variable to specify (ex. `hasEnabledEvent(PROGRAM_START)` to get the enabled property of the gcode:start event)
- updated api to work with new data structure
	- also added clearAll api for support - deletes all events at once
- changed import function to delete all events before adding new ones
	- prevents old events from carrying over if they do not get overwritten by the new file
- bug fixes
---
all code that comes after events are in feederCB, so no changes were made there


kglovern:
One last thing - can you change the dev .rc file to edge_rc instead of sender_rc?  It's still shared between versions and if we alter the data structure in this it'll make it unusable if they use Main.

PR #241, opened 2022-12-09
Simplify Auto Diameter Probe Code

sophiabeluli:
- instead of setting wcs 0s, store location coordinates and do math


kglovern:
This was merged into the probe rework and so I'm closing this pull request - the changes will be available there for the auto-zero profile

PR #243, opened 2022-12-12
Optional Overlay Flag For Wizard Substeps

sophiabeluli:
- substeps have optional flag for overlay so user can't use widgets
- warning above the wizard to inform the user that widgets are disabled


walidkayhan:
- On the minimum screen size a part of the Widgets Disabled message is cut off and hidden behind the main wizard modal for some reason, could just be the way the absolute positioning is being handled?
Screenshot 2022-12-15 at 11 06 31 AM

- When on larger screen sizes, the Widgets Disabled message seems to move far away from the wizard modal, would prefer it to always be right on top of it with just a bit of spacing between them.
Screenshot 2022-12-15 at 11 06 49 AM

- Not a big one, but having the message be the same width as the wizard modal would make it look a bit cleaner and noticeable in my opinion

PR #258, opened 2023-01-05
Category Filter for Shortcuts

sophiabeluli:
- added dropdown to sort keybindings by category


walidkayhan:
Screenshot 2023-01-11 at 2 01 55 PM
1. Scroll bar and area should only be for the shortcuts table, when scrolling currently the filter area is hidden
2. Change to "Categories"
3. Would be cool if the category options were color coded as well like how they are in the table column


Screenshot 2023-01-11 at 2 04 31 PM
4. Change to "No Shortcuts..."
5. Scrollbar should be hidden if there is no data in the table
6. Remove horizontal padding (label and dropdown should take up the full width)

Looks and works really well other than that :)

walidkayhan:
Also, some of the categories are empty for some reason on my end, it's not showing all of the shortcuts for some reason, you can see in the last screenshot I filtered to only have display shortcuts under "Coolant" and it didn't display anything. Even with no filtering the coolant shortcuts aren't showing up for some reason.

kglovern:
Could we get conflicts removed with Dev so this can be merged?

walidkayhan:
> Could we get conflicts removed with Dev so this can be merged?

Done

PR #273, opened 2023-02-07
Mi 925 dev add setting and logic for rotary axis step per mm

ssamal94:
### Changes:

1. New settings inside ‘General’ preference to manage A-Axis step per MM
2. Events and functions to manage setting updates


src/app/containers/Preferences/General/Aaxis.js
```diff
@@ -0,0 +1,59 @@
...
+import store from 'app/store';
+import ToggleSwitch from 'app/components/ToggleSwitch';
+import Tooltip from 'app/components/TooltipCustom/ToolTip';
+import Input from '../components/Input';
+import Fieldset from '../components/Fieldset';
+
+
+const Aaxis = () => {
+    const [addAAxisStepPerMM, setAddAAxisStepPerMM] = useState(false);
+    //TODO
```
walidkayhan, line 13: The value here needs to persist, so it should be saved and retrieved from the store

src/app/containers/Preferences/General/Aaxis.js
```diff
@@ -0,0 +1,59 @@
...
+            // G0 Y0
+
+            const mmMacro = ''; //macro content
+
+            //Run macro
+            controller.command('macro:rotaryMMenable', mmMacro);
+            return;
+        }
+        controller.command('macro:rotaryMMdisable');
+    };
```
walidkayhan, line 47: - Would prefer better naming conventions for the functions and variables, don't need to specify the MM. Ex. the function name "handleRotaryAxisMMToggle" can just be "**handleRotaryAxisToggle**"
- Comments here can just be removed, don't have to worry about that logic now, it could possibly not even on the front-end

src/app/containers/Preferences/General/Aaxis.js
```diff
@@ -0,0 +1,59 @@
...
+    return (
+        <Fieldset legend="A Axis">
```
walidkayhan, line 50: Should be labeled "Rotary Axis"

src/app/containers/Preferences/General/Aaxis.js
```diff
@@ -0,0 +1,59 @@
...
+            <Tooltip content="When ON, Y-axis max will match the work piece circumference of stock diameter entered below" location="default">
+                <ToggleSwitch
+                    label="A-Axis Step Per MM"
+                    checked={addAAxisStepPerMM}
+                    onChange={() => handleRotaryAxisMMToggle(addAAxisStepPerMM)}
+                    size="small"
+                    style={{ marginBottom: '1rem' }}
+                />
+            </Tooltip>
```
walidkayhan, line 61: - The point of this toggle is to just set a flag within the app so we know we are in a sort of "rotary axis" mode if that makes sense, so this tooltip text doesn't quite make sense in our scenario so it needs to be updated
- ToggleSwitch label needs to be "Enable Rotary Axis"

src/app/containers/Preferences/General/Aaxis.js
```diff
@@ -0,0 +1,59 @@
...
+            <Tooltip content="Set work piece diameter" location="default">
+                <Input
+                    label="Workpiece Diameter"
+                    units={store.get('workspace.units')}
+                    onChange={null}
+                    additionalProps={{ type: 'number', disabled: !addAAxisStepPerMM }}
+                    value={0}
+                />
+            </Tooltip>
```
walidkayhan, line 70: This would be called "*Stock Diameter*", not Workpiece Diameter

src/app/containers/Preferences/General/Aaxis.js
```diff
@@ -0,0 +1,59 @@
...
+                    checked={addAAxisStepPerMM}
+                    onChange={() => handleRotaryAxisMMToggle(addAAxisStepPerMM)}
+                    size="small"
+                    style={{ marginBottom: '1rem' }}
+                />
+            </Tooltip>
+            <Tooltip content="Set work piece diameter" location="default">
+                <Input
+                    label="Workpiece Diameter"
+                    units={store.get('workspace.units')}
```
walidkayhan, line 65: Retrieving the current workspace units needs to be handled differently here, the first time changing the units will not cause the value here to be updated.

src/app/containers/Preferences/General/index.jsx
```diff
@@ -23,6 +23,7 @@
 
 import React from 'react';
 import JoggingPresets from './JoggingPresets';
+import Aaxis from './Aaxis';
```
walidkayhan, line 26: Again would prefer a better naming convention, components included, this can be named "RotaryAxis". You would have to update it here and in the actual component itself

ssamal94:
**New changes:**

1. Variables, functions and event names changed.
2. 'units' is now fetched from the state instead of the store.
3. Tooltip messages and Rotary Axis toggle label updated.
4.  Function to handle the Rotary Axis toggle now has more clarity and is neat. 100% logic to be handled on the server.
5. Aaxis component was renamed to 'RotaryAxis'.
6. Commented logic to run Macro deleted from the server.

**Tests:**

1. Any new console or server errors? **None**
2. Toggle works as expected with the correct toast message.
3. Event to handle toggle on the server works and prints data from the front end.

PR #274, opened 2023-02-09
Mi 947 add shortcuts for new a axis

ssamal94:
On review with Walid

### DIFF
--- a/src/app/containers/Preferences/General/RotaryAxis.js
+++ b/src/app/containers/Preferences/General/RotaryAxis.js
@@ -3,9 +3,7 @@ import React, { useState } from 'react';
 import store from 'app/store';
 import ToggleSwitch from 'app/components/ToggleSwitch';
 import Tooltip from 'app/components/TooltipCustom/ToolTip';
-import { Toaster, TOASTER_SUCCESS, TOASTER_WARNING } from 'app/lib/toaster/ToasterLib';
-import get from 'lodash/get';
-import reduxStore from 'app/store/redux';
+import { Toaster, TOASTER_SUCCESS } from 'app/lib/toaster/ToasterLib';
 import Input from '../components/Input';
 import Fieldset from '../components/Fieldset';
 
@@ -15,20 +13,7 @@ const RotaryAxis = ({ state }) => {
     const { units } = state;
     //TODO
     const handleRotaryAxisToggle = () => {
-        const firmwareType = get(reduxStore.getState(), 'controller.type', 'grbl');
-        const isGrbl = firmwareType.toLocaleLowerCase() === 'grbl';
         const shouldEnableRotary = !rotaryAxisStatus;
-
-        //If grbl, turn off Rotary Axis and update user
-        if (isGrbl) {
-            store.set('rotaryAxisStatus', false);
-            Toaster.clear();
-            Toaster.pop({
-                type: TOASTER_WARNING,
-                msg: 'Your firmware does not support Rotary Axis.',
-            });
-            return;
-        }
 
         store.set('rotaryAxisStatus', shouldEnableRotary);
         setRotaryAxisStatus(shouldEnableRotary);
--- a/src/app/containers/Preferences/General/RotaryAxis.js
+++ b/src/app/containers/Preferences/General/RotaryAxis.js
@@ -3,7 +3,9 @@ import React, { useState } from 'react';
 import store from 'app/store';
 import ToggleSwitch from 'app/components/ToggleSwitch';
 import Tooltip from 'app/components/TooltipCustom/ToolTip';
-import { Toaster, TOASTER_SUCCESS } from 'app/lib/toaster/ToasterLib';
+import { Toaster, TOASTER_SUCCESS, TOASTER_WARNING } from 'app/lib/toaster/ToasterLib';
+import get from 'lodash/get';
+import reduxStore from 'app/store/redux';
 import Input from '../components/Input';
 import Fieldset from '../components/Fieldset';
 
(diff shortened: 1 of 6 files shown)


src/app/store/shortcuts.js
```diff
@@ -777,4 +778,40 @@ export const shortcuts = [
...
+        title: 'Jog: A-',
+        keys: ['ctrl', '4'].join('+'),
+        cmd: 'JOG',
+        payload: {
+            axis: { [AXIS_A]: -1 },
+        },
+        preventDefault: false,
+        isActive: true,
+        category: JOGGING_CATEGORY,
+    },
```
walidkayhan, line 804: These shortcuts need to be in the /app/JogControl/index.js, hence why they are having trouble rendering to the list of shortcuts. The logic for jogging using the shortcuts is also there, so you can use it to check if the a-axis values are being sent to the controller.

Also when adding new shortcuts, make sure the id is unique, i believe a few of them used here are already taken with other shortcuts. We'll probably change the way ids are set in the future but for now they need to be unique manually

src/app/widgets/Visualizer/index.jsx
```diff
@@ -934,6 +935,22 @@ class VisualizerWidget extends PureComponent {
...
+            // TODO - delete this and write logic
+            console.log('A - ', axis.a);
+        },
+        UPDATE_STATUS: (_, { command }) => {
+            // TODO - delete console log and fire this event
+            console.log('Controller command: ', command);
+            // controller.command(command, store.get('rotaryAxisStatus', false));
+        }
+    }
+
```
walidkayhan, line 953: Not necessary, like mentioned before the a axis jog shortcuts should be moved to where all the other jog shortcuts are, and use the same function they use to determine jogging

src/app/widgets/Visualizer/index.jsx
```diff
@@ -1277,8 +1294,47 @@ class VisualizerWidget extends PureComponent {
...
+            keys: ['ctrl', '5'].join('+'),
+            cmd: 'ROTARY_AXIS',
+            payload: {
+                command: 'rotaryAxis:updateState',
+            },
+            preventDefault: false,
+            isActive: true,
+            category: GENERAL_CATEGORY,
+            callback: this.rotaryAxisFunctions.UPDATE_STATUS
+        },
```
walidkayhan, line 1335: Not necessary to have add shortcuts here anymore as the useKeybinding hook only takes in values from the shuttleControlEvents

src/app/widgets/JogControl/index.jsx
```diff
@@ -572,6 +585,45 @@ class AxesWidget extends PureComponent {
...
+            keys: ['ctrl', '4'].join('+'),
+            cmd: 'JOG',
+            payload: {
+                axis: { [AXIS_A]: -1 },
+            },
+            preventDefault: false,
+            isActive: true,
+            category: JOGGING_CATEGORY,
+            callback: this.shuttleControlFunctions.JOG,
+        },
```
walidkayhan, line 613: The "cmd" property needs to be unique, that is why A- is not showing up on the list

walidkayhan (commented):
Jogging on the A-axis also needs to be accomodated in src/app/lib/combokeys.js like the other axes

ssamal94:
### Changes:

**Added the following shortcuts:**

1. **Rotary Axis** - turns Rotary Axis ON/OFF
2. **Jog: A+** - Turns the Rotary Axis in _positive_ direction
3. **Jog A-** - Turns the Rotary Axis in _negative_ direction
4. **Go To A Zero** - Turns the Rotary Axis to its origin/initial position
5. **Zero A Axis**  - Sets the current location of the Rotary Axis to zero

**Note**: There are just the function blueprints (No logic)

### Tests:

1. All 5 shortcuts appear on the shortcut list and can be edited.
2. Rotary Axis shortcut toggles the status and console logs on the server
3. All shortcuts  console log correct data from within the correct function
Screenshot 2023-02-09 at 1 24 38 PM
Screenshot 2023-02-09 at 1 26 39 PM

walidkayhan:
Screenshot 2023-02-09 at 1 46 29 PM

- Getting error 2 bad number format when trying to jog with the A-axis, the value being sent to it is null as per my screenshot. It needs to properly send the correct value like the other axes.

- Passed that as well, there needs to be a check somewhere where if the current firmware is grbl we cannot send A-axis commands to the controller.

ssamal94:
### New changes: 

1. Rotary Axis settings and events blocked for GRBL

src/app/widgets/Location/index.jsx
```diff
@@ -440,9 +448,16 @@ class LocationWidget extends PureComponent {
...
+            const rotaryAxisStatus = store.get('rotaryAxisStatus');
+            const firmwareType = get(reduxStore.getState(), 'controller.type', 'grbl');
+            const isGrbl = firmwareType.toLocaleLowerCase() === 'grbl';
             if (!axisList || axisList.length === 0 || activeState !== GRBL_ACTIVE_STATE_IDLE) {
                 return;
             }
+            if (axisList[0] === 'a' && !rotaryAxisStatus || axisList[0] === 'a' && rotaryAxisStatus && isGrbl) {
+                return;
+            }
+            console.log('LOCATION EVENT, Axis: ', axisList); // TODO - Delete this
```
walidkayhan, line 460: Please move this, going to be merging this PR soon

src/app/containers/Preferences/General/RotaryAxis.js
```diff
@@ -13,7 +15,21 @@ const RotaryAxis = ({ state }) => {
...
+        if (isGrbl) {
+            store.set('rotaryAxisStatus', false);
+            Toaster.clear();
+            Toaster.pop({
+                type: TOASTER_WARNING,
+                msg: 'Your firmware does not support Rotary Axis.',
+            });
+            return;
+        }
+
```
walidkayhan, line 32: The guard doesn't need to happen here, please remove all the firmware check logic here and all. The rest of the guards on the jog controls look good

PR #278, opened 2023-02-13
Shortcut Issues

sophiabeluli:
- fix some shortcut titles
- removed ids
    - created migration for removing ids
    - now uses cmd to compare and identify keybindings
- fixed bug where keybindings would fire during edit (ex pressing F4 opened the calibration modal)


walidkayhan (approved):
Looks good, will probably do a bit more refactoring in the future but its definitely a huge improvement over the previous iteration

PR #282, opened 2023-02-27
Regression With XY Squaring and Movement Tuning

sophiabeluli:
- change id for tabbedWidget component


Spikesef:
When I am calibrating movement on Gsender ver1.4.7, I get an NaN value.  What does this mean?

PR #285, opened 2023-03-01
Mi 969 dev unable to trigger probe multiple times in a row using probe widget

ssamal94:
### Changes:

1. Turning off the probe test allows probing even without touchplate
2. User can now probe multiple times in a row

### Tests:

1. Probe connectivity test toggled ON - The app waits for probe circuit confirmation. Users can't proceed with the probe.
2. Probe connectivity test toggled OFF - The 'Start Probe' button is available for the users.
3. Multiple Probe Test - Started probing for multiple axes, stopped probe in the middle of an operation, unlocked alarm; it allows to probe repeatedly without any issue.

### Errors and warnings:

- **None**

Screenshot 2023-03-01 at 7 27 15 AM
Screenshot 2023-03-01 at 7 28 20 AM

### DIFF
--- a/src/app/actions/preferencesActions.js
+++ b/src/app/actions/preferencesActions.js
@@ -6,17 +6,14 @@ export const {
     HOLD_SHORTCUTS,
     UNHOLD_SHORTCUTS,
     SET_IP_LIST,
-    SET_PROBE_CONNECTION_MADE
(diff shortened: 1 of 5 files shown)


ssamal94:
**New changes:**
Probe check state management moved out of redux

All tests are successful.

PR #306, opened 2023-03-16
Mi 940 setup conditional logic for displaying or enabling certain UI elements depending on firmware compatibility and loaded file

ssamal94:
### Changes:
When Rotary is ON, do the following things:
- If Firmware is Grbl and Rotary is ON = Disable Y and Enable A
- If Firmware is Grbl and Rotary is OFF = Disable A and Enable Y
- If Firmware is grblHal = Enable both A and Y
- Steps/mm, maximum rate, and acceleration firmware settings updated with fake values
- Constant file has default values stored
- If on Grbl, display a reminder modal every time Rotary Status changes to ensure the users if they changed their wiring over

**TODO:**
- FInd a way to know what axes are present in a file
- Use the above logic to complete the canStart() function in Visualizer to decide if we should start the file or not

Screenshot 2023-03-16 at 3 14 08 PM
Screenshot 2023-03-16 at 3 28 29 PM


src/app/lib/rotary.js
```diff
@@ -20,6 +26,13 @@ export const updateWorkspaceMode = (mode = WORKSPACE_MODE.DEFAULT) => {
         const prev101Value = controller.settings.settings.$101;
         store.replace('workspace.rotaryAxis.prev101Value', prev101Value);
         controller.command('gcode', `$101=${ROTARY_AXIS_101_VALUE}`);
+
+        if (firmwareType === 'Grbl') {
+            controller.command('gcode', `$103=${ROTARY_AXIS_103_VALUE}`);
+            controller.command('gcode', `$113=${ROTARY_AXIS_113_VALUE}`);
+            controller.command('gcode', `$123=${ROTARY_AXIS_123_VALUE}`);
```
walidkayhan, line 33: Seems unnecessary to invoke the command function three separate times, you can just pass it an array of values, ex. `controller.command('gcode', ['gcode1', gcode2' gcode3']`)

src/app/widgets/JogControl/Keypad.jsx
```diff
@@ -155,6 +159,10 @@ class Keypad extends PureComponent {
         const canClickXY = canClickX && canClickY;
         const canClickZ = canClick && _includes(axes, 'z');
 
+        const { ROTARY } = WORKSPACE_MODE;
+        const rotary = store.get('workspace.mode') === ROTARY;
+        const firmwareType = get(reduxStore.getState(), 'controller.type');
```
walidkayhan, line 164: Don't see the point of using `reduxStore.getState()` here, better practice would be to retrieve the value and send it as a prop in the connect HOC, for reference you can look at the component export in `src/app/widgets/Spindle/index.jsx`

src/app/widgets/Location/DisplayPanel.jsx
```diff
@@ -400,13 +421,15 @@ export default connect((store) => {
     const activeState = get(store, 'controller.state.status.activeState');
     const canHome = isConnected && [GRBL_ACTIVE_STATE_IDLE, GRBL_ACTIVE_STATE_ALARM].includes(activeState) && workflowState !== WORKFLOW_STATE_RUNNING;
     const mpos = get(store, 'controller.mpos');
+    const firmware = get(reduxStore.getState(), 'controller.type');
```
walidkayhan, line 424: Again no point in using `reduxStore.getState()` here, especially since this is in an actual connect HOC, you already have access to the redux state from the `store` variable here

src/app/widgets/Rotary/StepsToggle.js
```diff
@@ -3,34 +3,64 @@ import React, { useState } from 'react';
...
+import styles from './index.styl';
 
 const { DEFAULT, ROTARY } = WORKSPACE_MODE;
 
 const StepsToggle = () => {
     const [workspaceMode, setWorkspaceMode] = useState(store.get('workspace.mode', DEFAULT));
-
+    // eslint-disable-next-line no-unused-vars
+    const [wiringNotification, setWiringNotification] = useState(false);
+    const firmwareType = get(reduxStore.getState(), 'controller.type');
```
walidkayhan, line 20: Same drill, since this is a functional component you can use the `useSelector` hook to access the redux state easily

src/app/widgets/Rotary/index.styl
```diff
@@ -25,6 +25,10 @@
     overflow: auto;
 }
 
+.wiringMessage{
+    padding: 2rem;
+}
+
```
walidkayhan, line 31: Best practice for naming classes is to use kebab case so we can easily know that they are in fact css classes. Ex. `.wiring-message {`

walidkayhan (commented):
Mainly changes with how redux is used. Will do end-to-end testing once these changes are made

PR #324, opened 2023-04-18
Mi 1020 fix visualizer not diplaying the correct gcode shape for a file

ssamal2801:
### Changes:
The main change I made is to use a rotation matrix to perform the rotation instead of manually calculating the new x, y, and z coordinates. This should result in smoother curves since it's a more accurate and efficient way to perform rotations.

Here's a brief explanation of the changes I made:
1. First, I converted the angle from degrees to radians, since the makeRotationY and makeRotationZ methods expect the angle in radians.
2. Created a new Vector3 object using the vertex coordinates and set the z coordinate to 0 to match the orientation of the axis.
3. Created a new Matrix4 object using either the makeRotationY or makeRotationZ method, depending on the given axis.
4. Applied the rotation matrix to the point using the applyMatrix4 method and returned it.

By using a rotation matrix, this function should be more accurate and efficient at rotating the axis, resulting in smoother curves on the visualizer.

### Tests:

1. **Environment:**
-  Branch - MI-1020-Fix-visualizer-not-diplaying-the-correct-gcode-shape-for-a-file 
- macOS, web app
2. Ran rotary and non-rotary files
3. gSender VS NC Viewer:
Screenshot 2023-04-18 at 3 50 35 PM
Screenshot 2023-04-18 at 3 50 43 PM

5. Visualizations:

Screenshot 2023-04-18 at 4 27 27 PM
Screenshot 2023-04-18 at 4 42 18 PM
Screenshot 2023-04-18 at 4 24 34 PM
Screenshot 2023-04-18 at 4 24 24 PM


walidkayhan:
Looks good but there are a few things I noticed:

### Knight 3D Finish File
- Its shape is mostly correct with a few exceptions, the top and bottom of the shape needs aren't quite correct as it is supposed to be a circular shape there, but in our visualizer, it is in a panel shape for some reason. You can validate and have a better understanding of this miscalculated shape by comparing the visualizer to ncviewer:

**Expected**
Screenshot 2023-04-19 at 1 51 25 PM


**Output**
Screenshot 2023-04-19 at 1 50 51 PM


### When the File is Running
- The gcode lines in the visualizer need to be greyed out to indicate they have been processed when running the job, it currently doesn't seem to be doing that to both rotary and even regular files, so there is a regression with that. I would think it would've been an isolated issue specific to rotary files but even regular files don't grey out the gcode lines anymore for some reason.

PR #328, opened 2023-04-22
Create ability to access and use $# parameters in Macros

neilferreri:
It's been requested to be able to use the PRB parameters in macros for probing accuracy.
After making a $# query for the parameters, they can be accessed in a macro.
This could be useful for other parameters as well. Using stored G28 or G30 values, XYZ values from any WCS, or the TLO value in macros creates more possibilities.
I also added example variables to the pre-defined macro variables.
image

I had all kinds of issues with whitespace (extra spaces or line feeds) that had me guessing.
(This is my first pull request, so happy to modify, improve, or just go away)


kglovern:
Thanks for this - we'll take a look in the next week or so as I have time 👍

AndyCXL:
Fully support this initiative, having been one of the recent 'explore how to probe more accurately' posse of Macro writers who have concluded that the PRB: output from GRBL-HAL is really the only sound way to get the accuracy sought (otherwise axis deceleration has to be 'guessed' into how macros work out an exact position once the G38.x triggers).

PR #331, opened 2023-04-27
Go To Location Function

sophiabeluli:
- added go to location button and modal
    - user can choose relative or absolute move
    - A axis input included if rotary mode is on


AndyCXL:
Could I ask if 'A axis included if rotary mode is on' means the slider option on the rotary tab is 'on', or whether it means that GRBL-HAL is reporting an A axis whether or not rotary mode is 'on. My machine has 5 axes (XYYZA) permanently available, hence the question.
Sorry for the late inclusion, given that review has been requested

sophiabeluli:
@AndyCXL Yes, the switch in the rotary tab must be enabled for the A axis input to appear here.

PR #335, opened 2023-05-04
Mi 1028 fix bug with tabbed widgets breaking with spindle/laser toggled on

ssamal2801:
### Changes: 
(Final Edit)
1. Added a dropdown to the last tab to change through widgets.
2. Screen width more than 1280px - no dropdown
3. Screen width between 1248px and 1280px - Only Coolant in the dropdown
4. Screen width less than 1248px - Displays both Rotary and Cooland in the dropdown


ssamal94:
On review:

The 'Menu' tag with class .dropdown holds a child 'MenuItem'. The aim is to select an inbuild class (.rc-dropdown-menu-item-active) from the MenuItem npm package to apply custom styles. You can see that class if you inspect on devtools. 

The problem:
It works perfect if I throw in the styles in (.rc-dropdown-menu-item-active) inside a CSS file. However, if I try to target the class in scss inside .dropdown class, it does not work.

ssamal94:
Ok that's what I thought. I did try to import the .styl file like a css and it doesn't work.

src/app/widgets/MoreWidgets/Dropdown.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+    const handleTabSelect = (tab) => {
+        updateDropdownTab(tab.label);
+    };
+
+    const DropdownWrapper = styled.div``;
+    const Span = styled.span``;
+    const I = styled.i`
+        font-size: 2rem !important;
+    `;
```
walidkayhan, line 47: These can be in a separate file and then imported here, otherwise currently they will be re-created on every render

src/app/widgets/MoreWidgets/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+        break;
+    case 'Macros':
+        Render = Macro;
+        break;
+    case 'Spindle/Laser':
+        Render = Spindle;
+        break;
+    default:
+        break;
+    }
```
walidkayhan, line 58: A neat and cleaner trick you can do instead of using a switch case here is to utilize an object instead:

```
const Render = {
   "Coolant": Coolant,
   "Rotary": Rotary,
   "Console": Console,
   "Macros": Macros,
   "Spindle/Laser": Spindle,
}[currentDropdownTab] || Coolant
```

You have the object associating each component by a string name, and you can target the current one by using `[]` on that object and passing in the `currentDropdownTab` variable. Having the `|| Coolant` at the end is the same as having a default value for this in case the `currentDropdownTab` value doesn't map to anything for any reason

walidkayhan:
### Two weird things I noticed:

1. The active tab is changed to another one for some reason when decreasing the screen size

https://github.com/Sienci-Labs/gsender/assets/45078505/64adcce5-80cc-48c8-9e25-d861aeb13c92

---

2. When one of the ellipsis tabs is active and you decrease the screen size, the contents of that tab disappear. It looks like there is no active tab at that moment for some reason

https://github.com/Sienci-Labs/gsender/assets/45078505/f28cf3f2-f468-4e1a-9424-b6f1203f3b01

ssamal2801:
https://github.com/Sienci-Labs/gsender/assets/39333609/d4035665-ab8e-4d09-bd56-802525a156b5


https://github.com/Sienci-Labs/gsender/assets/39333609/d66ef547-5336-4e5a-8512-d83cc6027fd8

Fixed wrong highlights and disappearing content

ssamal2801:
Fixed wrong highlight bug

https://github.com/Sienci-Labs/gsender/assets/39333609/fd43c0a3-cd51-45cd-96e1-09ef517d3005

PR #336, opened 2023-05-08
MI-1047-Disable-A-axis-DRO-and-Jogging-when-a-job-is-running-or-paused

ssamal94:
### Changes:
Both A-axis DRO and Jogging are disabled on job run or job pause.

### Tests:
1. DRO and Jogging controls are available when:
- no files loaded
- file loaded but not yet run
- file stopped while running or on pause
2. Start a file, DRO and Jogging are disabled
3. Pause a file while running, DRO and Jogging stay disabled

Screenshot 2023-05-08 at 10 26 49 AM
Screenshot 2023-05-08 at 10 26 33 AM
Screenshot 2023-05-08 at 10 26 42 AM

### DIFF
--- a/src/app/widgets/Rotary/index.js
+++ b/src/app/widgets/Rotary/index.js
@@ -112,7 +112,6 @@ const Rotary = ({ active }) => {
         }
     }, [controllerState]);
 
-    console.log(isFileRunning);
     return (
         <Widget>
             <Widget.Content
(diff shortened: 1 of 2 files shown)


src/app/widgets/Rotary/index.js
walidkayhan, line 1: Rather than having useState and useEffect here, we can utilize redux and grab the controller's state and connection. Refer to the `canClick` function in `src/app/widgets/JogControl/index.js line 1362`, we can pretty much do the same there with the only difference being how we retrieve the workflow, isConnected, and activeState variables from redux, can probably use the useSelector hook to retrieve them

PR #364, opened 2023-06-09
MI-1099-MAIN-connect-from-remote-client-stops-job-in-main-client

ssamal2801:
### Changes:

- Added guard so that when a job is running on the server and a client reconnects, it only loads the file to the visualizer and not on the controller. That means a job won’t stop if you reconnect a client if it is running on the server. 

### Tests:

- Does not stop a job if you close the window and connect from a new window.
- It does not stop a job if you refresh the page.
- It does not stop a job if auto-connect is off and you connect a machine after refreshing a page or re-open it from a different browser.

https://github.com/Sienci-Labs/gsender/assets/39333609/3de7be1e-f377-42f7-ab79-0f0bd008310c

### DIFF
--- a/src/server/controllers/Grbl/GrblController.js
+++ b/src/server/controllers/Grbl/GrblController.js
@@ -1293,6 +1293,9 @@ class GrblController {
             },
             'gcode:load': () => {
                 let [name, gcode, context = {}, callback = noop] = args;
+                if (this.workflow.state === WORKFLOW_STATE_RUNNING) {
+                    return;
+                }
                 if (typeof context === 'function') {
                     callback = context;
                     context = {};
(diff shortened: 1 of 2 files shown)


kglovern:
This prevents loading a file, connecting, and then starting the job - the job never actually starts.  

Connecting, loading, and THEN starting works fine.  

I'd maybe simplify it to always controller.loadFile, which calls gcode:load, and inside gcode:load you can add a guard there to check if sender isRunning and return early instead of loading the file.  This prevents reloading if a job is running but doesn't break otherwise functional workflow.

ssamal2801:
Handled all the file load edge cases

https://github.com/Sienci-Labs/gsender/assets/39333609/861ce253-d596-4f6a-99ca-e254e3798c10

PR #365, opened 2023-06-14
Feeder Should Account for Sender Queue

sophiabeluli:
Main Issue
- replace feederCB in specific scenarios with a more robust solution
- two scenarios:
    - on start/resume
        - issue: need to use feeder for event code before sender should start again
        - solution: set bool in sender to indicate it is waiting. on feeder 'complete', run sender
    - on pause
        - issue: need to use feeder for event code, but only after sender has stopped
        - solution: set bool in feeder to indicate it is waiting. on sender 'hold', run feeder

Big Fixes
- fixed bug with pause event


kglovern:
This was solved alternatively with your changes to how we handle OK's between the sender and feeder so we'll go with that for now - closing.

PR #383, opened 2023-07-10
Visualizer Loses Anti-Aliasing on Window Resize

sophiabeluli:
- switch from FXAA composer to SMAA pass
Note: I couldn't replicate the issue myself, but I thought instead of adding another composer, maybe adding the anti-aliasing to the composer we already have will work better. Also, I left the fxaa composer code commented out in case this solution ends up being worse.


kglovern:
Fixed with visualizer changes in edge

PR #393, opened 2023-07-26
Display Active Firmware Modals In Diagnostics

sophiabeluli:
- add firmware modals section to Your Machine in Diagnostics Panel

### DIFF
--- a/src/app/containers/Calibration/Diagnostics/components/ProgramInfo.jsx
+++ b/src/app/containers/Calibration/Diagnostics/components/ProgramInfo.jsx
@@ -23,17 +23,57 @@
 
 import React from 'react';
 import store from 'app/store';
-import { get } from 'lodash';
+import { get, isEmpty } from 'lodash';
 import { connect } from 'react-redux';
 import cx from 'classnames';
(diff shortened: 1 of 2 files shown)


src/app/containers/Calibration/Diagnostics/components/ProgramInfo.jsx
walidkayhan, line 1: No need to make any changes here, just a friendly tip 😃, but a great way to map a variable to another one is to use an object:

```
// Current method
const determineMovement = () => {
    if (modals.motion === 'G0') {
        return 'Linear (non-extrusion)';
    } else if (modals.motion === 'G1') {
        return 'Linear (extrusion)';
    } else if (modals.motion === 'G2') {
        return 'Arc (Clockwise)';
    } else {
        return 'Arc (Counter-Clockwise)';
    }
};

// ✅ A more clean approach
const movementLabel = {
    'G0': 'Linear (non-extrusion)',
    'G1': 'Linear (extrusion)',
    'G2': 'Arc (Clockwise)',
}[modals.motion] ?? 'Arc (Counter-Clockwise)';
```

PR #410, opened 2023-09-21
Send Machine Profile to Metrics

sophiabeluli:


walidkayhan:
This looks good, I will merge this in after the API is updated, which I'm currently working on. Merging it before will cause the entire data collecting feature to stop working...

PR #442, opened 2024-01-04
New Global Rounding + Store Value Format Change + Important Bug Fixes

sophiabeluli:
### Rounding
- made the rounding function be used in store.set so anytime smth is changed in store it will round before saving
    - have a list of valid keys, so only the store properties in that list will be rounded
    - function also recursively checks objects, so you can still store.set an entire object instead of each individual property
- added rounding to ControlledNumberInput with an option you can pass to it to not round (used for things like ip range)

### New Store Value Format
- created a migration for the store values that saved both mm and in so they only hold metric values now
- went through the code and converted all usages of these properties so it is compatible with the new format

### Bug Fixes
- fixed an issue with the merge function so rounded values are kept saved and default data is not corrupted


walidkayhan:
Looks good overall, one thing I noticed is when my preferred units are imperial and I refresh the page it converts the jog preset values again to imperial, even though they are already in imperial. This causes the preset values to eventually all be 0, xy move, z move, and speed.

PR #443, opened 2024-01-09
EEPROM Settings Not Applied

sophiabeluli:
- stopped unrelated error pop up in firmware settings
- added new error popup with details for settings that could not be applied


kglovern:
Changed base to Master for 1.4.x

PR #448, opened 2024-01-16
Stripping comments from gcode before forwarding to machines.

gsrunion:


crpalmer:
The comment matching should be done in two separate steps.  The commentMatcher is used to preserve the line comment for tool changes, pauses, etc.  So it should match and save /;.*/.  After that comment has been removed, the (...) comments can then be removed.

The regex that you are using for the (...) comments is incorrect.  It should be /\([^\)]*\)/gm;.  The command

X (comment) Y (comment ( ( ( comment) Z

should be X Y Z when you're done removing comments.  Your regex would incorrectly make it X Z.

If this were my code repository, I would want to see this as 2 comments.  The first changing the comments and the second adding in the extra regex.  You could also consider adding a filtering step which removes all the (...) comments before returning line at the end of Feeder and Sender.

gsrunion:
> The comment matching should be done in two separate steps. The commentMatcher is used to preserve the line comment for tool changes, pauses, etc. So it should match and save /;.*/. After that comment has been removed, the (...) comments can then be removed.
> 
> The regex that you are using for the (...) comments is incorrect. It should be /([^\)]*)/gm;. The command
> 
> X (comment) Y (comment ( ( ( comment) Z
> 
> should be X Y Z when you're done removing comments. Your regex would incorrectly make it X Z.
> 
> If this were my code repository, I would want to see this as 2 comments. The first changing the comments and the second adding in the extra regex. You could also consider adding a filtering step which removes all the (...) comments before returning line at the end of Feeder and Sender.

Ok. I'll look into making those changes. JS is not my strong suite, but I'll manage.

src/server/controllers/Grbl/GrblController.js
```diff
@@ -373,7 +373,8 @@ class GrblController {
                     });
                 }
 
-                return line;
+                // remove all comments before forwarding on
+                return line.replace(/([^\)]*)/gm, '').replace(commentMatcher, '');
```
GibsT, line 377: Change this to return line.replace(/\([^\)]*\)/gm, '').trim();

src/server/controllers/Grbl/GrblController.js
```diff
@@ -540,7 +541,8 @@ class GrblController {
                  * End of Rotary Logic
                  */
 
-                return line;
+                // remove all comments before forwarding on
+                return line.replace(/([^\)]*)/gm, '').replace(commentMatcher, '');
```
GibsT, line 545: Change this to return line.replace(/\([^\)]*\)/gm, '').trim();

src/server/controllers/Grblhal/GrblHalController.js
```diff
@@ -336,7 +336,8 @@ class GrblHalController {
                     line = line.replace('M6', ';M6');
                 }
 
-                return line;
+                // remove all comments before forwarding on
+                return line.replace(/([^\)]*)/gm, '').replace(commentMatcher, '');
```
GibsT, line 340: Change this to return line.replace(/\([^\)]*\)/gm, '').trim();

src/server/controllers/Grblhal/GrblHalController.js
```diff
@@ -479,7 +480,8 @@ class GrblHalController {
                     line = line.replace(`${tool?.[0]}`, `(${tool?.[0]})`);
                 }
 
-                return line;
+                // remove all comments before forwarding on
+                return line.replace(/([^\)]*)/gm, '').replace(commentMatcher, '');
```
GibsT, line 484: Change this to return line.replace(/\([^\)]*\)/gm, '').trim();

GibsT (changes requested):
1. Regex is missing escape character for parenthesis. 
2. Using replace(commentMatcher, '') is redundant as it is already done previously.

src/server/controllers/Grbl/GrblController.js
gsrunion: https://github.com/Sienci-Labs/gsender/pull/448/commits/c609723ecffe93c6ce104906b192a6f8be2ffdea

src/server/controllers/Grbl/GrblController.js
gsrunion: https://github.com/Sienci-Labs/gsender/pull/448/commits/c609723ecffe93c6ce104906b192a6f8be2ffdea

src/server/controllers/Grblhal/GrblHalController.js
gsrunion: https://github.com/Sienci-Labs/gsender/pull/448/commits/c609723ecffe93c6ce104906b192a6f8be2ffdea

src/server/controllers/Grblhal/GrblHalController.js
gsrunion: https://github.com/Sienci-Labs/gsender/pull/448/commits/c609723ecffe93c6ce104906b192a6f8be2ffdea

PR #450, opened 2024-01-19
removing inline comments before forwarding gcode to the machine

gsrunion:


kglovern:
Thanks, we'll take a look when we're able to 👍

kglovern:
https://github.com/Sienci-Labs/gsender/pull/454 also solves this with the addition of a PID for Shapeoko so we'll likely accept that over this, thanks for your contributions!

PR #454, opened 2024-01-20
Fixes for SO5

GibsT:
Strips inline comments from being sent to the serial port which will lock up a SO5. SO5 PID was added to the valid PID list.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gSender",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "description": "Electron sender for GRBL based CNC machines",
   "author": {
     "name": "Sienci Labs <hi@sienci.com>",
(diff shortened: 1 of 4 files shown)


kglovern:
This is just the equivalent fix for bracket comments on Shapeoko firmware as https://github.com/Sienci-Labs/gsender/pull/450 with the PID added?  We'll likely merge this in for next build instead of that one.

GibsT:
> This is just the equivalent fix for bracket comments on Shapeoko firmware as #450 with the PID added? We'll likely merge this in for next build instead of that one.

Yes it is.

gsrunion:
Yep. I have been meaning to close my PR. When would these changes be available on a build, edge or otherwise?

kglovern:
This is live as of 1.4.1, thank you both for your contributions.

PR #455, opened 2024-01-25
Time Estimation Buffer

sophiabeluli:
- created queue and countdown interval to simulate countdown even when buffer is full
- made a lot of changes and bug fixes (mostly to do with js math being terrible)


walidkayhan:
I wasn't able to get the countdown when the buffer was immediately full for smaller files, is this the expected behavior?

PR #469, opened 2024-02-01
Rotary A to Y Conversion

sophiabeluli:
- added A jogging to list for continuous jog stop
- use aStep for A commands and added default values for it
- convert A to Y when sending command

### DIFF
--- a/src/app/widgets/JogControl/index.jsx
+++ b/src/app/widgets/JogControl/index.jsx
@@ -264,7 +264,7 @@ class AxesWidget extends PureComponent {
         jog: (params = {}) => {
             const { units } = this.state;
             const modal = (units === 'mm') ? 'G21' : 'G20';
-            const s = map(params, (value, letter) => ('' + letter.toUpperCase() + value)).join(' ');
+            const s = map(params, (value, letter) => ('' + (letter === 'a' || letter === 'A' ? 'Y' : letter.toUpperCase()) + value)).join(' ');
             const commands = [
                 `$J=${modal}G91 ` + s,
             ];
@@ -494,13 +494,11 @@ class AxesWidget extends PureComponent {
     shuttleControlFunctions = {
         JOG: (event, { axis = null }) => {
             const isInRotaryMode = store.get('workspace.mode', '') === WORKSPACE_MODE.ROTARY;
-            const firmwareType = this.props.type;
-            const isGrbl = firmwareType.toLocaleLowerCase() === 'grbl';
             if (event) {
                 preventDefault(event);
             }
 
-            if (isGrbl && axis.a && !isInRotaryMode) {
+            if (axis.a && !isInRotaryMode) {
                 return;
             }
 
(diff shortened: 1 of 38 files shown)


src/app/widgets/JogControl/index.jsx
```diff
@@ -256,7 +256,7 @@ class AxesWidget extends PureComponent {
         jog: (params = {}) => {
             const { units } = this.state;
             const modal = (units === METRIC_UNITS) ? 'G21' : 'G20';
-            const s = map(params, (value, letter) => ('' + letter.toUpperCase() + value)).join(' ');
+            const s = map(params, (value, letter) => ('' + (letter === 'a' || letter === 'A' ? 'Y' : letter.toUpperCase()) + value)).join(' ');
```
walidkayhan, line 259: Don't think you need to do a check here, you should probably just change the letter property value from params to whatever axis is needed before, when you call the jog function.

src/app/widgets/Rotary/Rotary.js
```diff
@@ -78,7 +78,7 @@ const Rotary = () => {
         },
         jog: (params = {}) => {
             const modal = 'G21';
-            const s = map(params, (value, letter) => ('' + letter.toUpperCase() + value)).join(' ');
+            const s = map(params, (value, letter) => ('' + (letter === 'a' || letter === 'A' ? 'Y' : letter.toUpperCase()) + value)).join(' ');
```
walidkayhan, line 81: Same drill here, change the letter property to the necessary axis letter when calling the function.

src/server/controllers/Grbl/GrblController.js
```diff
@@ -1952,7 +1952,7 @@ class GrblController {
                     axes.F = axes.F.toFixed(3);
                 }
 
-                const jogCommand = `$J=${unitModal}G91 ` + map(axes, (value, letter) => ('' + letter.toUpperCase() + value)).join(' ');
+                const jogCommand = `$J=${unitModal}G91 ` + map(axes, (value, letter) => ('' + (letter === 'a' || letter === 'A' ? 'Y' : letter.toUpperCase()) + value)).join(' ');
```
walidkayhan, line 1955: I don't think you need to change A to Y here. If handled correctly, A-axis movements will be changed to the Y-axis in the front end and there is no need to do a check here for it again.

src/server/controllers/Grblhal/GrblHalController.js
```diff
@@ -1957,7 +1957,7 @@ class GrblHalController {
                     axes.F = axes.F.toFixed(3);
                 }
 
-                const jogCommand = `$J=${unitModal}G91 ` + map(axes, (value, letter) => ('' + letter.toUpperCase() + value)).join(' ');
+                const jogCommand = `$J=${unitModal}G91 ` + map(axes, (value, letter) => ('' + (letter === 'a' || letter === 'A' ? 'Y' : letter.toUpperCase()) + value)).join(' ');
```
walidkayhan, line 1960: Same drill as the Grbl Controller

walidkayhan (commented):
Looks good, but the 4th axis functionality no longer works. When not in rotary mode and on the 4th axis mode, the A-axis should be utilized and it shouldn't be converted to the Y-axis.

When in rotary mode, that's when you need to convert A to Y.

walidkayhan:
The jog buttons on the UI still don't work as expected, this needs to be the case:

### When in Rotary Mode
- UI jogging, DRO, zeroing, and go to for the A-axis in the rotary tab should be converted to the Y-axis

### When not in 4th Axis Mode (Default)
- UI jogging, DRO, zeroing, and go to for the A-axis in the rotary tab should persist and not be translated

PR #470, opened 2024-02-07
Add Macros to Gamepad Shortcuts

sophiabeluli:
- added macros to gamepad
- save macros in gamepad context that updates when shortcut preferences are opened


walidkayhan:
Looks great! One thing I did notice initially is if you delete the macro that is set to a button, it is still attached to that button as a shortcut, and it sets the title as the macro ID since it doesn't exist anymore.

Screenshot 2024-02-08 at 10 28 43 AM

If a macro that has a button assigned to it is deleted, it should be cleared from that button as well.

PR #492, opened 2024-03-20
Yellow/Grey Buffer Lines Improvements

sophiabeluli:
- use estimated time countdown to predict where bit is and therefore where greyed lines should be


kglovern:
This looks good when starting a job normally, but when using start-from-line the behaviour is not as expected.

Test file: MB-3.1

Start from line: 75000 / 150500

You'll notice half the file becomes yellow lines (planned) but the grey lines start from the origin.

Expected behaviour is that the first 75000 lines would be greyed out and planned lines would start from there.

kglovern:
This crashes the renderer when a job is run twice in a row.

Tested using jsdc fixture.

image

PR #509, opened 2024-05-01
Include A Axis in Start From Line Calcs

sophiabeluli:
- add A axis to initial gcode sent before start from line
- modify and move GcodeToolpath file to backend


walidkayhan:
This looks great, but I think there's a regression with the zero a-axis button not rotating the g-code modal to its original angle anymore for some reason, once that's fixed I'll merge this branch in :)

sophiabeluli:
Turns out that's a bug on main, so I fixed it and another similar bug I found and pushed to main.

PR #511, opened 2024-05-14
Surfacing Spiral, Center Start: Issue With Moving Below Safe Height

sophiabeluli:
- put move to safe height by itself
- remove ramp and replace with z movement


walidkayhan (changes requested):
Looks great! Just do some cleanup on the code and remove the commented out logic if its not needed anymore, ill merge this once that's done :)

PR #536, opened 2024-08-19
Emit the `gcode:load` event

Billiam:
:wave: I maintain  a cncjs pendant ([cncjs-pendant-streamdeck](https://github.com/Billiam/cncjs-pendant-streamdeck)) and am looking to add support for gsender.

One feature missing from gsender (that is available in cncjs) is the `gcode:load` event. This allows rendering/animating the currently loaded gcode in the pendant, as well as displaying dimensions and extent.


Billiam:
Closing, since the events `file:load` and `file:unload` already appear to meet my needs here.

PR #551, opened 2024-09-21
Ignore USB vendor/product ID casing in device list

dymk:
On macOS, USB product/vendor IDs are lower-case hex encoded strings. This change implements case-insensitive comparison for product and vendor IDs, ensuring that recognized devices (such as ESP32) appear correctly under the "Recognized Devices" section.

A new `caseInsensitiveIncludes` function has been added to perform case-insensitive checks on array values. The port recognition logic now uses this function to compare product and vendor IDs, accommodating both upper and lower case hex strings.


kglovern:
Thanks, we'll take a look and likely merge this in 👍

PR #576, opened 2024-11-15
Allow M6 command to passthrough the firmware. 

siganberg:
The current passthrough switch flag isn't functioning correctly; it modifies the M6 command in a way that toolchange-supporting firmware can't interpret. It also triggers a confirmation dialog window, which isn't necessary for a functioning ATC.

https://www.youtube.com/watch?v=1Jvo0tx4MSM


siganberg:
Outdated PR.

PR #578, opened 2024-11-28
Firmware Detection on Connection

sophiabeluli:
- created a connection class that handles:
    - the initial connection and detection of the firmware
    - emitting serialport commands (ex. serialport:write)
    - sending/receiving commands from the serial connection
- once firmware type is found, the cnc engine creates the correct controller like normal
- the controller no longer deals with the connection directly, it uses the connection class


kglovern:
For context: COM5 = grbl, COM6 = grblHAL device

Some notes: 
- Should probably remove firmware select from connection widget if it's now redundant

- IP option should always appear in connection widget instead of being based on selected firmware type

- Multiple connect messages in console
image

- Controller type in redux inconsistently updates between grbl/grblHAL

- It looks like the grblController at least is always entering the "I need to hard reset to get a startup message" branch on connection which results in a slow experience
image

- Connected status is referencing the old selection value instead of the detected firmware flavour (connected to grbl, showing grblHAL which is what's selected)
image

- Ethernet connection chooses wrong controller (grbl) when it should default to grblHAL since that's the only flavour that supports ethernet connectivity.
image
image

- Ethernet also goes into slow reset block but that might just it using grbl instead of grblHAL.

- Failure to connect on Ethernet leads to a state where no connection is possible with any method (usb/ethernet)

walidkayhan:
Feedback:

- The firmware type shown in the connection widget is incorrect, though it does seem like it is choosing the correct firmware when connecting to the device
Screenshot 2024-11-29 at 2 06 13 PM
Screenshot 2024-11-29 at 2 06 05 PM

PR #605, opened 2025-02-24
Mi 1608 fix visualizer unmounting problem

sophiabeluli:
- replaced tanstack router with react-router
- routing is now done in `react-routes.tsx` file in `app/src/`
- server side rendering now uses a different component
- fixed some bugs and typescript errors


walidkayhan:
This PR requires additional testing, as it caused the front-end to show a blank screen, I'm assuming there's some issues with the new router implementation using react-router but ill have to investigate further

Issue #675, opened 2025-06-29
Minimum window width too wide for portrait mode on some monitors.


Japi42:
The 1.5.0 release of gSender changed the minimum window width to 1200 pixels ( [https://github.com/Sienci-Labs/gsender/commit/8a1cd7e4ad283e0cb96e4742b2dd113cfcf672df#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79L177-R232](url) ).  This causes portrait mode to be cut off on my 1680x1050 monitor.  Setting the minimum width back to 1024 (like was in Edge 8) appears to work fine (in Linux, not tested in MacOS or Windows).

PR #676, opened 2025-06-29
Change minimum window width back to 1024.

Japi42:
In gSender 1.5 the minimum window width was changed to 1200, which is too wide for portrait mode on most monitors (See issue #675) .  This change reverts the minimum width back to 1024 so portrait mode can be enjoyed once again.


SienciLabs:
@Japi42 thanks for the PR. We're still working on trying to get gSender to look good across different screen sizes and until we reach that point the minimum resolution will have to remain larger than 1.4.# versions. On a positive note though, I was able to make some changes for 1.5.4 that allowed us to reduce the minimum resolution down to 1040 x 796 which gets very close to the desired width to be used in portrait mode on most monitors. https://github.com/Sienci-Labs/gsender/commit/92c7be3bcf14e2195bb43bcb424226e21939fa24

Once we can get it a little bit further then this aspect should be back to 'fixed'! Just requires a little bit more playing around with design responsiveness. Once again thanks for the PR and bringing this to our attention, have a great day :)

PR #747, opened 2025-10-20
[MI-1849]: Navigation Blocker

sbeluli:
***don't merge today as we don't want this in today's build
Issue: https://sienci-labs.atlassian.net/browse/MI-1849
- added a navigation blocker
    - blocks navigation from the config page if there are unsaved changes
    - shows pop up that asks user if they are sure they want to continue with unsaved changes
        - if `yes`, continues navigating
        - if `no`, stops the navigation and stays on the config page
- made some typescript error fixes


ShilpaKAjin:
When a configuration value is modified and the user navigates away from the Config page without saving, a confirmation popup appears as expected. However, upon returning to the Config page, the unsaved value remains displayed, and the “Apply Settings” button remains active.

PR #788, opened 2026-01-21
fix(web): Fix iOS Safari file selection for G-code files

kesmit13:
## Summary
- iOS Safari shows G-code files (.gcode, .nc, etc.) as greyed out/unselectable in file picker dialogs
- This is because iOS Safari relies on MIME types rather than file extensions
- Added device detection for iOS/iPadOS and use MIME types (`text/plain`, `application/octet-stream`) for those devices
- Non-iOS devices continue to use extension-based filtering

## Test plan
- [x] Test on iPad Safari - verify .gcode files from Dropbox/iCloud are selectable
- [x] Test on desktop browsers - verify extension filtering still works
- [x] Test on iPhone Safari

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kesmit13:
The downside to this fix is that you can select any file, not just gcode files. However, without the fix you can't select anything, which I think is worse.

Issue #803, opened 2026-02-24
Updates to TLS tool change routine


kieraneglin:
Hey there! I have been playing around with TLS tool changing on my Altmill and I think there are some possible improvements that I wanted to gauge interest on before I create a PR:

### Tool presentation location 

Based on my workspace, my TLS has to be set up awkwardly near the back-right side of the machine. The issue is that the tool change wizard immediately goes to my TLS location which means I have to jog the machine to present the tool to me so I can change it before continuing with probing. 

I'd like to add an _optional_ setting that allows specifying a tool presentation location. Here's what I imagine the new workflow could look like: 

- M6 is encountered 
- the wizard opens, starting with the "Starting Off" page. Simultaneously, the spindle moves to the specified tool presentation location to make it easier to switch your bit
- progressing through the wizard works as it currently does. If using Fixed Tool Sensor, probing would move the spindle to the specified probing location. 

My use case is for the Fixed Tool Sensor wizard, but I reckon this is a useful option for any tool changing strategy besides Ignore. Since it's optional, the behavior would be entirely opt-in and backward compatible. 

### Improved initial tool behaviour 

This might be a "me" issue. I'm using Fusion for what it's worth. 

If I run gcode with tool changes, I'm immediately prompted to go through the tool changing wizard. This first "change" first probes the length of the initial tool, followed by the "new" tool. The issue is that a tool change hasn't actually occurred and probing the new tool is entirely unnecessary. 

Based on context, I'm guessing this initial tool prompt shouldn't happen until the first M6 _after_ some milling has occurred. My initial thought is to suppress the the second probe (or the wizard entirely) if the first M6 occurs before any M3/M4 commands. I'm open to ideas here! I can also make it another setting if you'd like (something like "ignore initial tool change")

---

Let me know what you think! I would like a little feedback on some of my points, but if you're interested I would be happy to put up some PRs (:

kglovern:
Partial duplicate of https://github.com/Sienci-Labs/gsender/issues/672 (tool change location)

Generally we're open to it with the caveat that there's not trickle down impact on the new ATC system.

I'll share my internal doc on requested/planned TC wizard updates here since some of it is relevant and it's been covered before (not all of this is relevant, just collected issues):

https://brash-elf-7d6.notion.site/Tool-change-Wizard-updates-30d79195bff080e8bbfcd8abd285ea13

**General**

[...]

Issue #672, opened 2025-06-25
Feature Request - Tool Change Location


gcormier:
For the "normal" toolchange mode, please allow us to define a toolchange location, and change wizard workflow such that the gantry moves to this new toolchange location and pauses. Once the dialog button is pushed on the screen, rapid to tool-setter location, measure tool.

This allows us to pick a nice toolchange location, such as the front middle of the machine.

Bonus points - allow us to define a feedrate for this move.

gcormier:
Park is used for feed-hold and raises the Z position clear and turns off my spindle. I would prefer keeping that functionality (raise Z, do not move X/Y when feedhold)

I see your consideration but would point out that for someone to use the fixed location toolsetter included in the current gSender, homing sensors are also required.

gcormier:
@walidkayhan Actually

`$346` is the setting.

<img width="958" height="304" alt="Image" src="https://github.com/user-attachments/assets/9b5d684e-52d4-447d-942b-752518593242" />

in gSender 1.5.6, this doesn't seem to work. It works in ioSender 2.0.46.

Is it because of the toolchange wizard? I will re-open.

PR #805, opened 2026-02-25
[FEAT] Add customizable manual tool change location

kieraneglin:
See #803, #672

## Preamble

I'm no stranger to managing open-source projects, so feel free to tear my approach apart. You won't hurt my feelings and I know how hard it is to keep things cohesive when you have random contributors like me

## What's new

- Adds two new options to settings:
    - `workspace.toolChange.moveToManualPosition` - a boolean that indicates whether the toolhead should move to an initial position for manual tool changing before moving on to the probing step. Defaults to `false` and only shows up for the `Fixed Tool Sensor` strategy
    - `workspace.toolChange.manualPosition` stores the aforementioned location
- Adds a new step to the `Fixed Tool Sensor` wizards
    - This step moves the toolhead to the specified location after raising to a safe Z
    - This step is conditional and only appears if `workspace.toolChange.moveToManualPosition` is set
    - ⚠️ NOTE ⚠️ I was not able to test this part. I don't own the CNC I work with so I can't run untested code on it, no matter how simple. I can include a dirt-simple gcode file with toolchanges if that'd make your life easier

## Notes

- Nothing much to add here, but I will have annotations on my code changes highlighting my thoughts and possible refactors that I didn't want to get into initially

### DIFF
--- a/src/app/src/wizards/automaticToolchange.tsx
+++ b/src/app/src/wizards/automaticToolchange.tsx
@@ -76,20 +76,36 @@ const probeInitialToolStep = [{
     ],
 }];
 
-const moveToToolchangePositionStep = [{
-    label: 'Move to Tool Change Location',
-    cb: () => {
-        const manualPosition = store.get('workspace.toolChange.manualPosition', { x: 0, y: 0, z: 0 });
-        controller.command('gcode', [
-            '(Moving to manual toolchange location)',
-            'G90 G53 G0 Z[global.toolchange.Z_SAFE_HEIGHT]',
-            `G90 G53 G0 X${manualPosition.x} Y${manualPosition.y}`,
-            `G90 G53 G0 Z${manualPosition.z}`,
-        ]);
-    },
-}]
+const moveToToolchangePositionSubstep = [{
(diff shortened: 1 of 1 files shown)


.prettierignore
```diff
@@ -1,2 +1,5 @@
 # Ignore server for prettier, different code styling
 src/server
+
+# TODO: remove
+**
```
kieraneglin: Will undo before merge. For some reason I can't stop prettier from formatting any file I touch

kglovern:
> > We are ultra busy wrapping up product launches for next month, but I will try to make time to look at this next week in office.  No promises however it will be run when I can.
> 
> Do what you need! I'm patient and I don't want to introduce additional testing burden 🤙🏻
> 
> Also, how do you test changes that require a CNC? Do you run it on an actual machine or do you have some sort of emulator for testing?

One of the benefits of working for a company that makes and sells CNC machines is I get access to CNCs - We have about 10 setup for testing/engineering work.  I'll run probably run it through a couple times in air and then with wood depending on any issues.

PR #806, opened 2026-02-25
[FIX] Update settings to be more reactive

kieraneglin:
## Preamble

I encountered a bug when setting up my tool length sensor where the probe coordinate input wouldn't appear when expected. Here are the repro steps that still work on Edge-4:

- Under `Tool Changing` in the settings, set `gSender Strategy` to `Pause` and save settings. Reload the page for good measure
- Back under `Tool Changing`, set `gSender Strategy` to `Fixed Tool Sensor` and save settings
- **Expected:** I would be shown the inputs to enter the XYZ coordinates of my probe immediately after saving
- **Actual:** The XYZ inputs don't appear until after I save AND toggle some other, unrelated setting (eg: toggling `Skip Dialog`)

## The fix

Super simple. I added `pubsub.publish('repopulate');` to the end of `Settings.ts#updateAllSettings`. I could instead add that line to an `onUpdate` callback on the `workspace.toolChangeOption` setting itself, but it seemed to me like this is something that should be called in all cases anyway

## What's new

While I was here, I also looked into improving reactivity of the form as a whole. Personally, I expected the XYZ inputs to appear as soon as I updated the strategy dropdown rather than requiring me to first save the form. To that end, I've updated `SettingsMenu.ts` and added a helper method to allow settings to react based on other setting's dirty values instead of just their persisted values.

It seems to work well but I don't know if this would have any unexpected knock-on effects. I only updated `workspace.toolChangePosition` as a minimal PoC but if you like the approach then I'll apply it to all conditional settings!

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,2 @@
 # Ignore server for prettier, different code styling
 src/server
-
-# TODO: remove
-**


kglovern:
Closing this because I implemented it in a slightly better way (no need for pubsub, we can just have a memoized map of dirty settings and retrieve them through a function that returns the pending value if it exists or the store value if not).

Same result, bit easier to maintain + uses the existing info we have.  This is available in dev, and should be applied to all relevant hidden functions.  This should fix both issues.

PR #809, opened 2026-02-27
Enhance accessibility features and keyboard support across UI

ashleycox:
I'm a totally blind maker learning CNC for the first time. I use a screenreader (specifically Voiceover on MacOS) which converts text on the screen to synthesised  speech. Many of the UI controls within GSender lacked labels that the screenreader could read, and therefore the interface was largely inaccessible. These contributions add descriptive aria labels to numerous UI controls, implement live regions, landmarks and roles, introduce accessibility settings such as reduced motion, keyboard focus indicators, focus trapping for modal dialogues, a context-sensitive keyboard shortcuts map, and desciptive g-code summaries for the visualiser. These additions make gSender the only fully accessible CNC control application around, as far as I am aware.


chrismakesstuf:
This is so cool @ashleycox! What an incredible contribution, we very much appreciate you putting this together ❤️ it should be ready to be tried out in 1.6.0 Edge 5 now!

PR #838, opened 2026-04-17
[FEAT] Add shutdown button for host computer in top bar

kieraneglin:
This one is really hyper specific to my needs so I totally understand if this doesn't make the cut! Here's the short version of why I wanted this:

- We have an AltMill 4x4 that's used by multiple people. Most of them are older and not super computer savvy
- We use the gControl Panel Computer in portrait mode. Because of the above tech literacy issue, we hide a lot of the Windows UI so all the guys can see is gSender
- We've had to disable the silver power button on the gControl since Windows no longer allows showing a confirmation prompt before shutdown. Our only options when pressing the silver power button are to shutdown immediately without prompting or to ignore the input entirely. We chose the latter since guys were accidentally turning off the computer when trying to insert a USB stick

Because of those factors, it would be really great to have a way to shut down the host computer via gSender. Again, I totally get that this is a very specific use case so feel free to tell me "no" 😉

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -31,6 +31,7 @@ import {
     session,
     clipboard,
 } from 'electron';
+import { exec } from 'child_process';
 import { autoUpdater } from 'electron-updater';
 import Store from 'electron-store';
 import chalk from 'chalk';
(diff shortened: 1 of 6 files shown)


src/main.js
```diff
@@ -515,6 +516,23 @@ const main = () => {
...
 
+            ipcMain.on('shutdown-host', () => {
+                let command;
+                if (process.platform === 'win32') {
+                    command = 'shutdown /s /t 0';
+                } else if (process.platform === 'darwin') {
+                    command = 'osascript -e \'tell app "System Events" to shut down\'';
+                } else {
+                    // Linux, including Raspberry Pi
+                    command = 'shutdown -h now';
```
kieraneglin, line 527: This requires sudo permissions and will fail without them. No way around that, unfortunately

kglovern:
Sorry, way too niche.

PR #839, opened 2026-04-17
[FIX] Update toolchange positioning mode

kieraneglin:
This change is related to [this comment/thread](https://forum.sienci.com/t/fixed-location-tool-change-issue-soft-limits/9421/56?u=keglin) on the forums.

I've gone through the auto toolchange/probe tool length routines and identified some potential issues. I've highlighted those along with some questions in `automaticToolchange.tsx`. Any comments on that file also apply to `probeToolLength.tsx` but the files are so similar that I didn't want to repeat myself

---

⚠️ I have only tested this in a grblHAL sim. I have not tested this on a real machine ⚠️


src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -21,25 +21,20 @@
...
+	const tcZPos = Math.abs(tcPosition.z);
 
-	//const curZPos = Math.abs(Number(get(state, 'controller.mpos.z')));
-	const position = store.get("workspace.toolChangePosition");
-	const curZPos = Math.abs(position.z);
-
-	return (maxZTravel - curZPos - 2).toFixed(3);
+	// leave 2mm of buffer to ensure we don't hit the max travel limit and cause a hard stop
+	return (maxZTravel - tcZPos - 2).toFixed(3);
 };
```
kieraneglin, line 38: - `_zProbeDistance` was unused so I removed it
- Minor renaming of variables for clarity but otherwise no functional changes

src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -57,20 +52,21 @@ const probeInitialToolStep = [
 						label: "Probe Initial Tool",
 						cb: () => {
 							controller.command("gcode", [
-								"G91 G21",
+								"G90 G21", // Set absolute positioning (modal)
```
kieraneglin, line 55: This is the core of the change and, really, I'm not sure how this worked before. My best guess is that the `G53` operated in absolute mode regardless of the `G91`.

If my reading of the spec is correct then `G53` should only work in absolute mode, hence this change

src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -57,20 +52,21 @@ const probeInitialToolStep = [
...
 								"G49", // cancel applied TLO offsets
 								"G53 G0 Z[global.toolchange.Z_SAFE_HEIGHT]",
 								"G53 G0 X[global.toolchange.PROBE_POS_X] Y[global.toolchange.PROBE_POS_Y]",
 								"G53 G0 Z[global.toolchange.PROBE_POS_Z]",
+								"G91 G21", // Set incremental positioning (modal) just for probing moves
 								"G38.2 Z-[global.toolchange.PROBE_DISTANCE] F[global.toolchange.PROBE_FEEDRATE]",
-								"G0 Z[global.toolchange.RETRACT]",
-								"G38.2 Z-10 F[global.toolchange.PROBE_SLOW_FEEDRATE]",
+								"G0 Z[global.toolchange.RETRACT_DISTANCE]",
+								"G38.2 Z-[global.toolchange.ADVANCE_DISTANCE] F[global.toolchange.PROBE_SLOW_FEEDRATE]",
```
kieraneglin, line 63: - I've renamed `RETRACT` to `RETRACT_DISTANCE` for clarity
- I've added `ADVANCE_DISTANCE` (which you'll see defined later in this file). It's the `RETRACT_DISTANCE` plus a small offset. Basing the slow probe distance off the `RETRACT_DISTANCE` makes things a little safer and also ensures everything continues to work even if `RETRACT_DISTANCE` is changed to be a larger number. I haven't tested, but I'm willing to bet that the old approach wouldn't work if you had set your `RETRACT_DISTANCE` in settings to 20mm for some reason

src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -125,22 +122,17 @@ const createWizard = (count: number) => {
...
 			const position = store.get("workspace.toolChangePosition");
-			// Get $13 value for adjustment of Z Safe Height
-			const state = reduxStore.getState();
-			const $13 = get(state, "controller.settings.settings.$13", "0");
-			const zSafe = $13 === "1" ? "-0.5" : "-10";
-
-			const zProbeDistance = calculateMaxZProbeDistance(
-				settings.zProbeDistance,
-			);
+			const zSafe = "-10"; // millimeters
```
kieraneglin, line 125: The old code that referenced `$13` and whatever else seems like it had a bug in it but I could be wrong.

My read is that this file would use `zSafe`/`Z_SAFE_HEIGHT` only after changing the units to millimetres. I couldn't find a case in which `Z_SAFE_HEIGHT` would be used while still in the user's preferred unit modal. Given that, it seems to me like `zSafe` can be assumed to only ever be used in the context of millimetres.

I haven't tested but I suspect that running the old toolchange code on a machine set to inches would set the Z safe height to `-0.5` _millimetres_ which isn't the end of the world but it doesn't seem intended

src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -125,22 +122,17 @@ const createWizard = (count: number) => {
...
 
 			controller.command("gcode", [
 				"%wait",
 				`%global.toolchange.PROBE_THICKNESS_MM=${settings.zProbeThickness}`,
 				`%global.toolchange.PROBE_DISTANCE=${zProbeDistance}`,
 				`%global.toolchange.PROBE_FEEDRATE=${settings.fastSpeed}`,
 				`%global.toolchange.PROBE_SLOW_FEEDRATE=${settings.slowSpeed}`,
-				`%global.toolchange.RETRACT=${settings.retract}`,
+				`%global.toolchange.RETRACT_DISTANCE=${settings.retract}`,
+				`%global.toolchange.ADVANCE_DISTANCE=${settings.retract + 2}`,
```
kieraneglin, line 135: Highlighting for viz

src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -223,14 +214,13 @@ const createWizard = (count: number) => {
...
 								label: "Resume Cutting",
 								cb: () => {
-									const unit = getUnitModal();
 									controller.command("gcode", [
 										"(Returning to initial position)",
 										"G53 G0 Z[global.toolchange.Z_SAFE_HEIGHT]",
-										`G90 ${unit} G0 X[global.toolchange.XPOS] Y[global.toolchange.YPOS]`,
-										`G90 ${unit} G0 Z[global.toolchange.ZPOS]`,
+										`G90 [global.toolchange.UNITS] G0 X[global.toolchange.XPOS] Y[global.toolchange.YPOS]`,
+										`G90 [global.toolchange.UNITS] G0 Z[global.toolchange.ZPOS]`,
```
kieraneglin, line 221: I couldn't see why `const unit = getUnitModal();` would be preferred over `[global.toolchange.UNITS]`. My reasoning is that `global.toolchange.XPOS` (etc) is going to be set based on the gcode file's defined unit at the time the toolchange starts so we probably want to restore based on that rather than what the user has defined in their settings.

I could be extremely wrong here so please let me know!

src/app/src/wizards/automaticToolchange.tsx
```diff
@@ -223,14 +214,13 @@ const createWizard = (count: number) => {
...
 										"(Restore initial modals)",
-										"M3 [global.toolchange.UNITS] [global.toolchange.DISTANCE] [global.toolchange.FEEDRATE]",
+										"[global.toolchange.SPINDLE] [global.toolchange.UNITS] [global.toolchange.DISTANCE] [global.toolchange.FEEDRATE]",
```
kieraneglin, line 223: Minor, but I wanted to add this in case an `M3` was incorrect and this should restore to an `M4` or whatever.

Again, if this change misses some context then please let me know and I'll revert it

kieraneglin:
@kglovern / @walidkayhan this is now ready for review! Thanks for your patience 🤙

src/app/src/wizards/automaticToolchange.tsx
kglovern: G53 is machine position so de facto always absolute and non-modal.

kieraneglin: In my research for this I read somewhere that running G53 while in G91 was undefined behaviour/implementation-specific but the more I look into it the more I think that person was incorrect.

I think it is good to change it anyway for clarity but it's looking like this line specifically has low/no practical impact

kglovern: Yeah, not sure about that either - most CNC standards actually label G53 explicitly as "Move in Absolute Machine coordianates".  Regardless, it's explicitly defined behaviour in the grbl/grblHAL implementations 👍

PR #844, opened 2026-04-20
[FEAT] Pass non-realtime console commands through feeder

kieraneglin:
I'm toying around with another idea and this popped up as something that could be helpful. I decided to scope it as its own PR to keep things simple.

**What's new?**

- When using the console in the Carve UI, non-realtime commands are now sent via `controller.command` so they can use variables, etc
    - NOTE: Realtime commands (`~`, `!`, etc) are passed immediately and unchanged
- Commands that are sent through the console have a new label of `feeder (console)`. This visually separates user-run manual commands from other commands fed through the feeder
    - More generally, commands can indicate their source in the console by using a `source` key in the provided context
- Realtime commands are now immediately echoed back in the terminal so the user has confirmation their command has been passed through
- The `feeder` prefix in the terminal has had its colour changed to one with much higher contrast

**Screenshots**

Screenshot 2026-04-20 at 11 28 23 AM


src/app/src/features/Console/TerminalInput.tsx
```diff
@@ -10,6 +10,8 @@ import { addToInputHistory } from 'app/store/redux/slices/console.slice';
 import { useTypedSelector } from 'app/hooks/useTypedSelector';
 import controller from 'app/lib/controller';
 import { toast } from 'app/lib/toaster';
+import { GRBL_REALTIME_COMMANDS } from 'server/controllers/Grbl/constants';
+import { GRBLHAL_REALTIME_COMMANDS } from 'server/controllers/Grblhal/constants';
```
kieraneglin, line 14: Let me know if you don't want to be including server files in the UI. I figured it was better than adding duplicate realtime commands to the frontend constants file, but I can be convinced otherwise

src/app/tsconfig.json
```diff
@@ -40,6 +40,9 @@
             ],
             "app-root/*": [
                 "./*"
+            ],
+            "server/*": [
+                "../server/*"
```
kieraneglin, line 45: This was needed to import server files in the UI. There was already an eslint alias for this which was surprising

src/server/services/cncengine/CNCEngine.js
```diff
@@ -535,7 +535,7 @@ class CNCEngine {
             });
 
             socket.on('command', (port, cmd, ...args) => {
-                log.debug(`socket.command("${port}", "${cmd}"): id=${socket.id}`);
+                log.debug(`socket.command("${port}", "${cmd}"): id=${socket.id}, args=${JSON.stringify(args)}`);
```
kieraneglin: This was just for my testing - I can remove it if you'd like or I can leave it if it provides value 🤙

src/app/src/features/Console/TerminalInput.tsx
```diff
@@ -37,7 +48,25 @@ const TerminalInput = ({ onClear }: Props) => {
...
+            controller.command('gcode', command, {
+                source: 'feeder (console)',
+            });
+
+            if (command.startsWith('%')) {
+                // The feeder normally eats commands that start with % but
+                // we still want to show them in the terminal for user feedback
+                onWrite(command, 'feeder (console)');
+            }
+        }
```
kieraneglin, line 69: This is the meat of this change

src/app/src/features/Console/TerminalInput.tsx
walidkayhan: We prefer the two to be separate, if you'd like you can create a common folder that both the frontend and server directories have access to

src/app/tsconfig.json
walidkayhan: Again, would like to separate the app and server as much as possible.

src/app/src/features/Console/index.tsx
```diff
@@ -135,7 +135,7 @@ const Console = ({ isActive, isChildWindow }: ConsoleProps) => {
             </div>
             <div className="grid grid-rows-[1fr_auto] absolute gap-1 top-0 left-0 w-full h-full p-1">
                 <Terminal ref={terminalRef} isActive={isActive} />
-                <TerminalInput onClear={handleTerminalClear} />
+                <TerminalInput onClear={handleTerminalClear} onWrite={(data, source) => terminalRef.current?.write(data, source)} />
```
walidkayhan, line 138: Any particular reason why the terminal write was moved to the main Console component? Not wrong per se, just curious

walidkayhan (commented):
Looks good, I have a couple of comments here. Some merge conflicts also need to be addressed.

PR #866, opened 2026-05-06
[FEAT] Dynamically generate ramping zigzag pattern for surfacing gcode generator

kieraneglin:
I was surfacing some parts yesterday and I noticed that the ramp angle of 10º was really aggressive for my 40mm surfacing bit. Caused a god-awful noise and some friction burns on the wood surface.

This PR updates that logic so that a ramp angle of 2º is used. To account for surfaces of different dimensions, a zigzag pattern is generated where the number of back-and-forth movements (legs) is determined by the size of the surfacing area.

---

I used a different approach in my first commit. Initially, I would adjust the ramp angle between 2º and 10º depending on the length of the surfacing operation but the zigzag approach works better on smaller pieces


src/app/src/features/Surfacing/utils/surfacingGcodeGenerator.js
```diff
@@ -63,8 +62,8 @@ export default class Generator {
 			"G90",
 			`${spindle} S${spindleRPM}`,
 			...dwell,
-			"G0 X0 Y0",
 			`G0 Z${z}`,
+			"G0 X0 Y0",
```
kieraneglin, line 66: NOTE: I also moved this so it goes to Z first. I can see cases for either but I reckon this approach is the one that's less likely to result in a crash.

I can easily be swayed - let me know if you want me to revert this 🤙

PR #879, opened 2026-06-01
Feature/toolhead outline

lykowdk:
Build a black edge "outline" of the tool model — the geometric edges of
the bit (its flutes and silhouette) drawn as black lines on top of the
mesh. Added as a child of the rotating cuttingTool so the outlined flutes
sweep as it spins, instead of the bit reading as a featureless grey blob.
EdgesGeometry keeps only edges whose adjacent faces meet more sharply than
thresholdAngleDeg, so smooth round areas stay clean and the flute creases
stand out. Lower the threshold to catch more edges, raise it for fewer.


lykowdk:
image

kglovern:
Sorry, not particularly interested in adding this.

PR #880, opened 2026-06-01
Add run-progress toolpath coloring to the 3D visualizer

lykowdk:
While a job runs, color the toolpath by execution state instead of a single grey/yellow split: GREY behind the real toolhead (executed), ORANGE for a short look-ahead lead hugging the tool (buffered, about to be cut), and YELLOW for the rest of the controller's buffered range. Beyond the buffer the path keeps its planned color, so a long file is never flooded.

GCodeVisualizer.js:
- greyOutLines(currentLineRunning, cutterPos) rewritten as a three-zone painter. The grey/orange split is found by a nearest-SEGMENT scan over the in-flight range using clamped point-to-segment distance to the machine position — robust on back-and-forth raster paths (the tool lies on its own segment, so it wins over a spatially-near parallel pass). The floor is monotonic so the executed edge never retreats; duplicated joint vertices are skipped.
- Crisp boundary overlay: a short depth-test-off line with a temp vertex at the machine position, so the grey/lead split is sharp under the marker rather than a gradient across the active segment.
- Orange lead sized by ~20mm of path travel via a cached cumulative arc-length table.
- The split only tracks while the toolhead is near the path; jogging off the path freezes it at the last cut point. Within a segment the split is monotonic and never slides backward (tool jitter / a small jog-back keeps the progress already made).
- setFrameIndex/render/unload reset the progress trackers and arc-length cache; unload disposes the overlay.

Visualizer.jsx:
- Pass the reported work position into greyOutLines so the split anchors at the on-screen cutter.

### DIFF
--- a/src/app/src/features/Visualizer/GCodeVisualizer.js
+++ b/src/app/src/features/Visualizer/GCodeVisualizer.js
@@ -34,6 +34,14 @@ const STATES = {
     DONE: 2,
 };
 
+// Lead color: buffered toolpath just ahead of the toolhead, between the
+// executed (grey) path and the rest of the buffered (yellow) path.
+const PROGRESS_LEAD_COLOR = '#ff8000'; // orange
+
+// Max distance (mm) from the toolhead to the nearest in-flight segment for the
+// split to track it; beyond this (jogging off the path) the split freezes.
+const PROGRESS_NEAR_MM = 2;
+
 class GCodeVisualizer {
     constructor(theme) {
         this.group = new THREE.Object3D();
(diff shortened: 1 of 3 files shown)


lykowdk:
image

walidkayhan:
We require a few unit tests for pull requests now. This particular one has a few new functions added, as well as some changes to current ones. If you can add unit tests to cover the changes, I can review this PR again and try to merge it in :)

kglovern:
We are reworking the visualizer as part of edge and this will be wrapped into the gviewer package separately.

PR #884, opened 2026-06-06
Add camera "follow tool" mode to the 3D visualizer

lykowdk:
A crosshair toggle (next to lite-mode) pan-follows the toolhead, keeping it centered while preserving the user's viewing angle and zoom. It eases smoothly toward the tool each animation frame (rAF loop, decoupled from the choppy status-report rate) and turns itself off when the user drags the view.

- index.tsx: followTool state (persisted via WidgetConfig) + camera.toggleFollowTool
- Visualizer.jsx: startCameraFollow() rAF easing loop; follow target = workPosition - pivotPoint so it works even when the tool STL model is null; drag-only disable via canvas pointer listeners (click/zoom keep following)
- definitions.ts / PrimaryVisualizer.tsx / VisualizerWrapper.jsx: types, the crosshair toggle button, and prop threading

### DIFF
--- a/src/app/src/features/Visualizer/PrimaryVisualizer.tsx
+++ b/src/app/src/features/Visualizer/PrimaryVisualizer.tsx
@@ -11,7 +11,7 @@ import Rendering from './Rendering';
 import SoftLimitsWarningArea from './SoftLimitsWarningArea';
 import CameraDisplay from './CameraDisplay/CameraDisplay';
 import { WorkspaceSelector } from 'app/features/WorkspaceSelector/index.tsx';
-import { FaFeatherAlt } from 'react-icons/fa';
+import { FaFeatherAlt, FaCrosshairs } from 'react-icons/fa';
 import cx from 'classnames';
 import { Tooltip } from 'app/components/Tooltip';
 import GcodeEditor from './GcodeEditor';
@@ -101,6 +101,7 @@ const PrimaryVisualizer = ({
                         <VisualizerWrapper
                             show={showVisualizer}
                             cameraPosition={cameraPosition}
+                            followTool={state.followTool}
                             ref={visualizerRef}
                             state={state}
                             actions={actions}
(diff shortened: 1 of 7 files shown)


walidkayhan:
We require at least a few unit tests for pull requests now, just to ensure the new logic that is requested to be merged into the master branch is as bulletproof as possible. For this change, adding unit tests for the new function would be necessary. Please add a few to this branch, and I will have a look at it again :)

lykowdk:
Good call — done! I've added 17 unit tests in `src/app/src/features/Visualizer/tests/cameraFollow.test.js`. To make them possible without mounting a full Three.js visualizer, I also extracted the four core methods into a standalone `cameraFollowHelpers.js` module (pure functions that take a context object), with `Visualizer.jsx` delegating to them.

The tests cover:
- `isFollowEnabled` — 6 cases: direct prop, nested `state.followTool`, falsy values, both set
- `getToolScenePosition` — 4 cases: with/without STL model loaded, missing `workPosition` fields, null `workPosition`
- `startCameraFollow` / `stopCameraFollow` — 7 cases: schedules rAF when enabled, no-ops when disabled or already running, cancels correctly, tick self-gates when toggled off mid-frame, tick settles when camera catches up

All 82 suite tests pass (`npm test`).

lykowdk:
The Linux and Raspberry Pi CI failures are **not caused by this PR**.

Both failures are in native module / Docker build infrastructure:

- **Linux** – `node-gyp` failed to rebuild the `usb` C++ addon during `electron-rebuild`, and `icon-tool.js` crashed processing the `.icns` icon. Neither touches any JS/TS code.
- **Pi** – Docker container is missing `libnspr4.so` (a shared library required by Electron), plus the same `icon-tool.js` crash.

This PR only modifies files under `src/app/src/features/Visualizer/` (pure JavaScript/TypeScript/JSX). It cannot affect native module compilation or Docker environment dependencies.

For reference, the most recent CI runs on `master` are all green — the same Linux and Pi jobs pass there — confirming these are transient infrastructure failures, not a regression introduced here.

The newest CI run for this branch is currently `action_required` (awaiting maintainer approval to run CI on a fork PR). Once approved it should pass all four platforms.

kglovern:
This is a baseline functionality in the edge visualizer and no longer required to be merged into Main.

PR #890, opened 2026-06-23
Add "Move To Here" click-to-move in the visualizer

gcormier:
image

Adds a "move to here" button. Intent to easily jog around based on the loaded file. My use case - find spots where I can safely drill some screws into my spoilboard to hold down a workpiece based on avoiding toolpaths :)

Goals were
1. Avoid accidental clicks
2. Touch and mouse friendly interaction

Note that the retract Z currently does not go back down. Personally I hate it any time my Z goes down without my consent (or running a job). Need feedback on Z behavior, as I know we also want to target machines without a homing sensor for gSender.

### DIFF
--- a/src/app/src/features/DRO/utils/SafeMove.ts
+++ b/src/app/src/features/DRO/utils/SafeMove.ts
@@ -1,0 +1,53 @@
+import controller from 'app/lib/controller';
+import store from 'app/store';
+import get from 'lodash/get';
+
+/**
+ * Build the gcode lines required to retract Z to the user's configured safe
+ * height before performing a rapid XY move.
+ *
+ * Mirrors the retract logic used by "Go To Location" so the two features stay
+ * consistent:
+ *  - When homing is enabled, retract in machine space (G53) but only if the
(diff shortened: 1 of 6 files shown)


gcormier:
@kglovern any thoughts on Z behavior?

kglovern:
> @kglovern any thoughts on Z behavior?

I'm fine with Z not going back down - that's how we treat safe height for go to movements without limit switches anyways currently.  Being consistent across homed/non-homed is fine.

The sticking point rn is we are in the midst of re-writing the visualizer completely since it's one of the last js files still in the front-end and largely CNCjs still so this would need to be migrated over if we were to add it as a feature.

gcormier:
I can close this then and re-open once you are done that - is that work on a branch right now?

kglovern:
> I can close this then and re-open once you are done that - is that work on a branch right now?

https://github.com/Sienci-Labs/gsender/tree/features/visualizer_update

It should be largely functional as is and I don't see the structure changing much, just a few edge cases to clean up

PR #906, opened 2026-07-13
WIP: Surface an unapplied tool length offset in the DRO

gcormier:
## The problem

After a fresh power-on, grblHAL zeroes the applied tool length offset. The tool table survives in NVS, and with `$485` the current tool number does too, but the *applied* offset does not: `gc_init()` memsets the parser state on a cold start, leaving G49 active.

The result is a controller that reports `T:1` with a perfectly good length sitting in tool table slot 1, while measuring from the spindle gauge line. Nothing re-applies the offset on its own. On my machine the Z DRO read 14.635mm short until I re-probed the tool, because the probe macro ends with a `G43` and that was the only thing putting the offset back.

This is not a display-only issue. The machine moves wrong too. Anything that just folds the TLO into the readout would make the DRO lie in the other direction.

## Why gSender could not see it

`GRBL_MODAL_GROUPS` and `GRBL_HAL_MODAL_GROUPS` had no group for the tool length offset words. `GrblLineParserResultParserState` / `GrblHalLineParserResultParserState` match each `$G` word against those tables and `continue` on a miss, so `G43`, `G43.1`, `G43.2` and `G49` were parsed and then discarded. gSender had the tool table, the current tool and the TLO parameter all in hand and still had no way to know the offset was not applied.

## Changes

- Add a `tlo` modal group to both controllers (`G43`/`G43.1`/`G43.2`/`G49` for grblHAL, `G43.1`/`G49` for grbl). These words were already being received and thrown away; they now populate `controller.modal.tlo`.
- Widen the `TLO` type to match what grblHAL actually reports (`report.c` emits `G43`, `G43.1` and `G43.2`, not just `G43.1`).
- Add a `ToolOffsetWarning` banner to the DRO, backed by a `useToolOffsetApplied` hook.
- Fix the `$G` echo in `GrblHalController.write()`. `replyParserState` was compared against `GRBLHAL_REALTIME_COMMANDS.GCODE_REPORT`, which is `'$G\n'`, but `cmd` has already been trimmed, so the comparison could never be true and a manually typed `$G` printed nothing at all. The grbl controller uses a literal `'$G'` and was unaffected; the grblHAL side regressed when the literal was swapped for the constant.

## When the warning fires

Only when every one of these holds:

- controller is grblHAL
- the machine has homed
- `T:` reports a tool above zero
- that tool's table entry has a non-zero Z offset
- `modal.tlo` is a `G49` that has actually been observed in a `[GC:]` report

No G-code is sent. It is a warning only.

## Deliberately not doing

**No automatic `G43`.** The real fix belongs in firmware, as a startup line (`$N0=G43` re-applies the persisted tool's offset after every successful `$H`, since a bare `G43` defaults its H word to the current tool). Having gSender inject the command would paper over exactly the kind of missing configuration this is meant to reveal.

`modal.tlo` is intentionally left undefined until a real parser-state report arrives, rather than defaulting to `G49`. Every path where we do not actually know resolves to no warning: no `[GC:]` reply yet, `G43.1`/`G43.2` dynamic offsets that the tool table knows nothing about, plain grbl, unhomed, tool 0, or a zero/unreadable table entry.

## Testing status

Untested on hardware beyond the diagnosis itself. Marking as draft.

### DIFF
--- a/src/app/src/features/DRO/component/ToolOffsetWarning.tsx
+++ b/src/app/src/features/DRO/component/ToolOffsetWarning.tsx
@@ -1,0 +1,41 @@
+import { AlertTriangle } from 'lucide-react';
+
+import {
+    Tooltip,
+    TooltipContent,
+    TooltipProvider,
+    TooltipTrigger,
+} from 'app/components/shadcn/Tooltip';
+
+import { useToolOffsetApplied } from '../utils/useToolOffsetApplied';
+
+export function ToolOffsetWarning() {
+    const toolOffsetApplied = useToolOffsetApplied();
+
+    if (toolOffsetApplied) {
+        return null;
+    }
+
+    return (
+        <TooltipProvider delayDuration={300}>
+            <Tooltip>
+                <TooltipTrigger asChild>
+                    <div
+                        className="flex flex-row gap-1 items-center justify-center py-0.5 text-xs font-semibold text-amber-700 dark:text-amber-500 cursor-default"
+                        role="status"
+                    >
+                        <AlertTriangle className="w-3 h-3 shrink-0" />
+                        <span>Tool offset not applied</span>
+                    </div>
+                </TooltipTrigger>
+                <TooltipContent className="max-w-72">
+                    The loaded tool has a length in the tool table, but the
+                    controller is in G49 so the work Z below does not include
+                    it. Run a tool change or send G43 to apply it, and set a
+                    startup line ($N0=G43) to have it re-applied automatically
+                    after homing.
+                </TooltipContent>
+            </Tooltip>
+        </TooltipProvider>
+    );
+}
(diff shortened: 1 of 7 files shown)


gcormier:
Leaving this as a draft for now. I want to run it on my own machine across a few power cycles and tool changes before I'd call it ready, in particular to confirm the warning clears the moment a `G43` lands and that it stays quiet during dynamic-TLO probing. I'll keep an eye on it and will un-draft once I've tested more.

gcormier:
Setting a `$N0` can negate the need for this change, but that seems like bad UX when our sender could warn us. I think the Sienci ATC sets $N0 as part of the bundled install, which is why maybe it was not noticed.

kglovern:
ATC sets N0 and N1 as part of installation (P100 and P200)

https://github.com/Sienci-Labs/atc_macros/blob/ef4c60f0ee3dac45af9241d7696a073a262028df/P999.macro#L52

Theoretically this could be simplified into a single macro (since P100 is just settings variables it's easy to prepend to the content of P200) but there's possible legacy installation issues that might arise if we did that

gcormier:
> ATC sets N0 and N1 as part of installation (P100 and P200)
> 
> https://github.com/Sienci-Labs/atc_macros/blob/ef4c60f0ee3dac45af9241d7696a073a262028df/P999.macro#L52

Should that be mandatory for gSender users with non Sienci ATC? Or should we surface this potential gap and tell the user?

PR #916, opened 2026-08-06
Portrait macro bar with drag-and-drop reordering

J-eremy:
I have recently upgraded my cnc computer and added a touch screen that I have set in portrait mode. While gsender works great in landscape mode I found myself fat fingering the wrong macro's more than I would like to admit. I added a macro bar that is disabled by default but has an option in the normal settings to turn it on. It only will show up if it is in portrait aspect ration AND there are macros.

## Summary

Adds a quick-access macro bar at the bottom of the screen when using gSender in portrait orientation. This is particularly useful for touch-screen or tablet setups on CNC machines where portrait mode is the primary layout.

## What changed

- **New `PortraitMacroBar` component** — renders a grid of macro buttons at the bottom of the Carve page in portrait mode, with drag-and-drop reordering via `@dnd-kit`
- **Carve layout updated** — bottom section now uses `flex-col` so the macro bar occupies space naturally without hardcoded padding
- **Toggle setting** — users can enable/disable the macro bar from Settings > Basic > UI Settings > "Portrait macro bar"
- **TypeScript interface** — `portraitMacroBar` added to the `Workspace` type
image

## How to test

1. Enable the setting in Settings > Basic > UI Settings > "Portrait macro bar"
2. Rotate to portrait orientation (or use a narrow viewport)
3. The macro bar appears at the bottom with your configured macros
4. Drag macros to reorder — the order persists across sessions
5. Disable the setting — bar disappears, no leftover padding
image

## Notes

- Follows the same patterns as `spindleFunctions` and `coolantFunctions` for the toggle (`useWorkspaceState`)
- When disabled, the bar takes no space — no absolute positioning or padding hacks
- Landscape mode is unaffected

image

### DIFF
--- a/src/app/src/features/Config/assets/SettingsMenu.ts
+++ b/src/app/src/features/Config/assets/SettingsMenu.ts
@@ -380,6 +380,14 @@ export const SettingsMenu: SettingsMenuSection[] = [
                         min: 0,
                         max: 4,
                     },
+                    {
+                        label: 'Portrait macro bar',
+                        key: 'workspace.portraitMacroBar',
+                        description:
+                            'Show a quick-access macro bar at the bottom of the screen in portrait mode.',
+                        type: 'boolean',
+                        defaultValue: false,
+                    },
                 ],
             },
             {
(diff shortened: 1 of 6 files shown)


kglovern:
There were some unintentional knock-on effects on the landscape mode view even with this option disabled but I've fixed them locally and we should be merging this.

PR #922, opened 2026-08-20
New Plugin Features

sbeluli:
- add gcormier's plugin PR
- add support for his changes
- add new sample plugin to demonstrate the new features


ShilpaKAjin:
1.The points marked in the visualizer is moving beyond the canvas of the visualizer 
2.The dropdown to change the workspace is hidden when 
3.The + button is not active when quick locate is clicked , it is taking more time than expected 
4.The distance between two points remains as same for mm and inches 
5.The “+” icon for marking points appears by default when reopening the Corner Finder. It should only appear after clicking Quick Locate. The Quick Locate button itself remains clickable.

sbeluli:
This is just a sample plugin to show off the new features gcormier added so I didn't change a few things bc it was unnecessary (as this isn't a real plugin people will be using):
- (2) dropdown hiding workspace
- (4) distance remains the same for mm and inches

ShilpaKAjin:
Bug to fix : The green markers are visible on top of Job status pop-up also G-code editor, start from line.

Verified Cases. 
Confirm com.sienci.corner-finder appears in the Plugin Manager with name "Corner Finder", version 0.1.0, and description text matching gsender-plugin.json.- PASS

The  plugin's toggle icon renders as the 🎯 emoji (single emoji, not text) in the visualizer-overlay slot.-- PASS

Confirm the plugin only becomes clickable/available when requiresIdle: true is satisfied ( Idle) 
--PASS

Click and mark count is correct  -pass

Distacne between last 2 points is calculated right  -pass

Camera preset buttons tested on lock rotation checked and un checked state -pass

Screen⇄world round-trip — button disabled with no pick yet -- pass

Confirm overlay markers clear from the main visualizer when you close the Corner Finder panel -- pass

Check round-trip button becomes enable when there is at least one marker  -- pass

ShilpaKAjin:
Bug fix verified and ready to merge. 

@walidkayhan @kglovern

PR #927, opened 2026-08-26
Features/remote mode recommended ports

kglovern:
## Description

The Remote Mode dialog now makes it easier to choose the right network address.

Previously, the Addr dropdown showed every IP address on the computer without much context. This made it easy to accidentally choose something like `127.0.0.1`, a virtual adapter, or an old network address that other devices could not use to connect to gSender.

Changes include:

* **Recommended address.** gSender now picks the address most likely to work for other devices on the same network, marks it as "Recommended", and selects it by default if no address has been saved yet.
* **Clearer address list.** Addresses now include a Wi-Fi or Ethernet icon and a short description. Hovering an address shows the adapter name.
* **Less useful addresses moved out of the way.** Loopback, virtual adapters, VPNs, and disconnected adapters are grouped under "Other / advanced" and greyed out. They can still be selected if needed.
* **Warning for addresses that may not work.** Selecting one of these addresses shows a warning explaining that other devices may not be able to reach gSender.
* **Handles old saved addresses.** If a previously saved address is no longer available, the dialog shows a warning and provides a link to switch to the recommended address.
* **Layout cleanup.** Addr and Port are aligned, the address dropdown has more room for descriptions, and the dialog stays the same height when changing addresses.

Saving settings and the restart prompt work the same as before.

## Type of Change

* [x] New feature
* [x] Refactor

## How Has This Been Tested?

* [x] Manually tested in Electron on Windows, in light and dark mode

## QA Instructions

1. Pull the branch, run `yarn dev`, and open the Remote Mode dialog.
2. **First open with nothing saved:** Addr should default to your active Wi-Fi or Ethernet address, not `127.0.0.1`. It should have a green check on the right. Confirm the address matches the active network shown by `ipconfig`.
3. **Open the dropdown:** normal Wi-Fi and Ethernet addresses should appear first. One should have a green "Recommended" label. Loopback, virtual adapters, and VPN addresses should appear below an "Other / advanced" divider and be greyed out. Hover over an address and confirm the adapter name appears.
4. **Pick an address that is unlikely to work**, such as `127.0.0.1` or a virtual adapter. The field should show an amber warning icon and warning text underneath. Switch between addresses and confirm the dialog does not resize.
5. **Test a saved address that disappears:** save Remote Mode using a Wi-Fi address, disconnect from that network, then reopen the dialog. An amber warning should explain that the saved address is no longer available and provide a link to use the recommended address. Click the link and confirm the recommended address is selected.
6. **Test the full connection:** enable Remote Mode, save, restart when prompted, then open the address or scan the QR code from a phone on the same network. Confirm the pendant loads.
7. **Also check:** dark mode looks correct, long addresses display properly, and a computer without internet access still shows available local addresses in the dropdown.

image
image


ShilpaKAjin:
Verified & Ready to merge 

- Remote Mode opens with the active Wi-Fi/Ethernet address and shows a green check
- Local addresses match `ipconfig` and the recommended address is labelled correctly
- Adapter names appear on hover
- Amber warnings display correctly
- Verified network disconnected state and confirmed the recommended-address link works.
- Tested QR code and pendant loaded successfully
- Verified dark mode, long addresses, and offline local-address detection

@kglovern @walidkayhan

PR #931, opened 2026-09-01
Features/sdk events response parsing

kglovern:
Plugins can now define their own parsers for firmware responses instead of
being limited to the hardcoded chain in the grbl/grblHAL runners. Parsers are
declared in the manifest or registered at runtime, matched server-side, and
delivered to the plugin as events.

## What's in here

- New parser engine in `src/server/lib/plugin-parsers/`: spec compiler, regex
  safety screen, `PluginParserChain` (line mode, block mode, capture windows,
  timeouts, rate limiting, quarantine).
- Wired into `GrblRunner` and `GrblHalRunner` via each controller's
  `runner.on("raw")` listener, which fires before any built-in parsing — so
  plugins see lines `serialport:read` never emits (status, json, descriptions,
  alarm/group details).
- Observe-only. Nothing a plugin registers can consume a line, alter built-in
  parsing, change state tracking, or touch the console feed.
- Manifest `parsers` array is read by pluginregistry and loaded on port open,
  so a plugin's parsers run whether or not its UI is mounted.
  `machine.registerParser` covers dynamic patterns at runtime, scoped to the
  registering iframe and replayed on reconnect.
- New SDK surface: `machine.registerParser`, `unregisterParser`, `onParsed`,
  `onLine`, `query`, `getLastParsed`, `onParserError`, and a `useParsed` hook.
- New `parser` bridge topic: a per-plugin last-value snapshot (for `useParsed`
  and late-mounting views) plus a lossless event stream (for `onParsed`).
- Two new permissions, `machine:parse` and `machine:query`, separate from
  `machine:read` — a parser reads the raw stream through its own regex, a
  wider grant than the curated `machine:read` context object. Import dialog
  lists every manifest-declared pattern verbatim.
- `plugins/parser-demo`, exercising every path with a command sender to drive it.

## Functionality tested

Automated, 136 tests passing:

- `src/server/lib/plugin-parsers/__tests__/`: spec validation, `g`/`y` flag
  stripping, caps/clamps, ReDoS rejection, block termination by
  `end`/`until`/`maxLines`/`timeout`, interleaved status reports ignored
  without terminating, `strict`, `restartOnBegin`, `whenWorkflow` gating,
  `reset("close")` flushing partials, rate limiting, quarantine, `feed()`
  never throwing.
- Regex safety screen checked against all 32 regexes gSender's own
  grbl/grblHAL parsers ship (all accepted) and known catastrophic patterns
  (all rejected) — guards against the screen over-rejecting ordinary
  firmware patterns.
- `pluginBridge.test.ts`: capability gating on the new request types and
  `parser` topic, per-plugin snapshot isolation, owner-id per iframe mount,
  runtime parser release on unmount, regression guard that `machine:command`
  never routes through `controller.command`.
- `capabilities.test.ts`: new scan-to-grant mappings, manifest parser grant
  merge, drift guard on the wire-type unions duplicated between `types.ts`
  and the SDK's `bridge.ts`.

Manual, without hardware: replayed a realistic 40-setting `$$` dump with
status reports interleaved at the real 250ms cadence through a live
`PluginParserChain` — one block, `complete: true`, 40 entries, status lines
excluded. Every manifest example in the docs was run through the real
validator.

Not tested against a machine yet — that's what the QA list below is for.

## How to QA

Build and install the demo:

```bash
cd plugins/parser-demo && npm install && npm run build
```

Run `npm run electron:dev` (dev loads `plugins/` directly, no copying
needed), connect to a machine, open **Tools → Parser Demo**.

On import, the permissions dialog should list `machine:parse` and
`machine:query` and show both declared patterns verbatim.

| # | Do this | Expect |
|---|---------|--------|
| 1 | Type `$$`, press Send | Returns promptly, not after 30s. Settings section fills with the whole dump, `complete: true`, no `<...>` status lines in it |
| 2 | Watch the console during 1 | Command and response appear as normal, no raw firehose, no duplicated `error:` lines |
| 3 | Send `$#` | Probe section fills from the `[PRB:]` line with named groups, field is `lastProbeTriggered` (not `succeeded` — renamed since `$#` reports the last probe cycle's outcome, not command success, and the old name read as a failure) |
| 4 | Query `$I`, default `until` | Resolves with build info, `reason: "until"` |
| 5 | Query with `until` = regex that never matches, timeout 2000 | Comes back after ~2s, `complete: false`, `reason: "timeout"`, does not hang |
| 6 | Navigate away from the plugin page, come back, send `$$` | Results still arrive — confirms manifest parsers keep running with no UI mounted |
| 7 | Disable the plugin in Tools → Plugins, send `$$`, re-enable | Matches stop, then resume, no app restart needed |
| 8 | Start a job, then query | Rejected as busy rather than interleaving |
| 9 | Disconnect mid-`$$` | Subscriber gets `complete: false`, `reason: "close"` rather than hanging |
| 10 | Register a runtime pattern in the demo's box, reconnect | Still matches after reconnect |
| 11 | Register `(a+)+` at runtime | Rejected with a reason, not accepted |
| 12 | Register `^` at runtime | Rate limited at 20/sec with a `plugin:parser:error`. Jogging and job streaming stay smooth |


ShilpaKAjin:
Verified all the 12 cases listed 
Also tested console copy bug fix and fix on  Fix height on dialog modals vs gcode editor vs visualizer camera cube
 
 Ready to merge 
 
 @kglovern @walidkayhan

PR #942, opened 2026-09-08
MI-2504

sbeluli:
## Description
- change park/go to/corner to use `pulloff distance` instead of `-1` if the bit for `set machine origin to 0` is not set

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Hotfix
- [ ] Refactor

## How Has This Been Tested?
- [x] Unit tests added/updated
- [ ] Cypress tests added/updated
- [ ] Manually tested in Chrome
- [ ] Manually tested in Safari
- [x] Manually tested in Firefox

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,8 @@ cypress/reports/
 cypress/Reports/
 cypress/downloads/
 
+plugins/**/ui/*
+packages/plugin-sdk/dist
+
+/.claude
+CLAUDE.md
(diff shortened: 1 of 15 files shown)


ShilpaKAjin:
The fix works correctly across all 4 affected spots 
Park, Go To, Go To Corner, and Manual Tool Change Position. All 27 of the PR's existing unit tests pass.
$27 = 0 → sends Z0 (no retract at all)
$27 negative (e.g. -5) → sends Z5 (wrong direction)
$27 non-numeric (e.g. "abc") → sends ZNaN (invalid G-code)
Full retract → XY → final-Z sequence for Manual Tool Change Position (was previously untested , only the retract line was checked)
Bug to fix 
$27 (pull-off distance) setting accepts non-numeric input, causing invalid ZNaN G-code

PR #943, opened 2026-09-09
Deduplicate Grbl/grblHAL controllers, step 1: shared line-parser result classes

CaliLuke:
## Description

Moves seven identical parser result classes used by the Grbl and grblHAL controllers into `src/server/controllers/shared/`:

- `Alarm`, `Echo`, `Error`, `Feedback`, `Help`, `Option`, and `Settings`

This is a refactor only. Parser behavior and registration order are unchanged. The remaining result classes stay separate because their implementations differ.

## Testing

This PR adds tests against the existing implementations before moving the classes. The test count in this area increases from 3 to 141.

The tests cover parser routing, payloads, registration order, status parsing, and runner state.

`npx jest --config jest.config.js` reports 366 passing tests, compared with 225 on `dev`. Three unrelated failures remain:

- `surfacing-output.test.ts` also fails on a clean `dev` checkout.
- Two `api.plugins` suites fail locally because Electron is not installed.

The server build succeeds with `node esbuild.config.js --target=server`. All touched files pass `biome format`.

This has not been tested on CNC hardware.

## Existing behavior preserved

The added tests exposed four differences between the parsers. This PR preserves them and adds tests for them:

- `[MSG:...]` returns `{message}` in Grbl and `{name, value}` in grblHAL.
- Grbl's `ok` matcher also accepts values such as `"o"` and `"k"`. grblHAL accepts only `"ok"`.
- The Grbl v0.9 `Lim:` branch discards its calculated pin state and checks bit 2 twice.
- grblHAL does not correctly parse v0.9 comma-delimited status reports.

These differences can be addressed separately after the team decides the intended behavior.

## Review notes

- The unused grblHAL `Option` class is removed instead of registered because registration would change behavior.
- Specific bracketed matchers still run before the generic `Feedback` matcher.
- `GrblHalLineParserResultAlarmDetails` remains separate because it is specific to grblHAL.
- `jest.config.js` now maps the `server/...` alias that esbuild previously resolved by itself.

## Type of change

- [x] Refactor

## Checklist

- [x] Relevant tests pass locally
- [x] Touched files pass `biome format`
- [ ] Tested with physical CNC hardware

## QA

This change has no expected visible effect.

1. Run `yarn dev` and connect a Grbl machine.
2. Make sure that alarms, errors, `[MSG:...]` feedback, and `$$` output still appear.
3. Trigger an alarm and make sure that the banner and message appear.
4. Run `$$` and make sure that the settings table loads.
5. Stream a short job and make sure that progress works as expected.
6. Repeat these steps with a grblHAL machine.


CaliLuke:
@kglovern could I get a review? my plan is to finish characterizing the Grbl/grblHAL core so it's safer to develop against it moving forward, this is the first step. Related to my old post https://forum.sienci.com/t/im-concerned-about-the-state-of-gsender/25291

kglovern:
Thanks, we saw this and will take a look as time is available!

CaliLuke:
is the PR size appropriate, should I go ahead and finish the job, should I make it smaller?

PR #953, opened 2026-09-17
Fix: don't report homing success or allow silent unlock after a failed homing cycle

fujiruki:
## Problem

When a homing cycle fails, Grbl reports ALARM 6-9 (homing fail / safety door / pull-off failure / limit switch not found). Two issues followed from treating this like any other alarm:

1. `GrblController`'s status handler set `hasHomedSet = true` on the next status report regardless of `activeState`, so a failed homing cycle could still be reported internally as "homed".
2. The unlock UI (`UnlockButton`, `MachineStatus`) sent `unlock` straight through for ALARM 6-9 with no warning, letting the operator jog or run G-code against a position that can no longer be trusted.

We hit this on a real machine: after `ALARM:8 (Homing fail)`, unlocking and jogging drove an axis into its limit switch and damaged it, because nothing indicated that homing had actually failed.

## Fix

- Only mark homing as successful when the status report after `$H` is not `Alarm`.
- Clear `hasHomedSet` whenever ALARM 6-9 is received.
- Ask for confirmation via the existing `ConfirmationDialog` before unlocking from ALARM 6-9 (`isHomingFailureAlarm` / `confirmUnlockAfterHomingFailure` in `UnlockButton`, reused from `MachineStatus`).

## Testing

- Added a unit test for the new `isHomingFailureAlarm` helper.
- `npx tsc --noEmit -p ./src/app` shows no new type errors from this change (pre-existing baseline errors excluded).
- Full Jest suite: only pre-existing, unrelated failures (plugin demos, JogStreamer, GcodeStepper, Surfacing, InstallPluginDialog); no failures in files touched by this change.

Happy to adjust naming/placement if this doesn't match the project's conventions.

https://claude.ai/code/session_013Y3nxXHDvZmr2wUhWYdozV


kglovern:
I think this is a good safety change, but homing failing now results in a loop where you can't home because the homing button is not available (disabled due to alarm) but clicking unlocks warns against unlocking after failed homing.

The main home either button needs a disabled state logic update to actually allow the user to rehome in alarm 6-9 or the primary action in the dialog needs to be to Home or unlock anyways instead of unlock or cancel if we want to add this safety feature.

Steps to recreate above:

1) Home on grbl board without motors attached
2) Click Unlock Machine alarm state button
3) Get prompted about making sure you want to unlock without rehoming without ability to send home command

fujiruki:
Thanks for flagging this — you're right that Home/Unlock both warning left the user with no clean way out.

Pushed a follow-up: the confirmation dialog now offers **Rehome** as the primary action (calls the same homing routine as the Home button) with **Unlock Anyway** as the secondary option, so re-homing is one click from the warning itself.

Additionally, for ALARM:8/9 specifically (limit switch not found / would not release — as opposed to 6/7), the dialog now explains that re-homing alone won't resolve it until the switch/wiring is fixed, and walks through disabling homing via $22 in Config so the machine can still be used while the switch is repaired.

